PR #24, opened 2020-03-28
BACKLOG-12935: reorerding of the server settings

fpral:
## JIRA

https://jira.jahia.org/browse/BACKLOG-12935

## Description

Reordered the server settings (here users and group entries) + removed user and groups icons in server administration (we need them only for site administration)

## Tests

The following are included in this PR
- [ ] Unit Tests (Most changes _should_ have unit tests)
- [ ] Integration Tests

## Checklist

I have considered the following implications of my change: 
it's just a reordering, nothing special here.

- [ ] Security (in particular for changes to authentication, authorization, data fetching, ...)
- [ ] Performance
- [ ] Migration
- [ ] Code maintainability

## Documentation

- [ ] Inline documentation
- [ ] Internal Documentation (wiki)
- [ ] User-facing Documentation

### DIFF
--- a/src/javascript/registrations/groups/registerRoutes.js
+++ b/src/javascript/registrations/groups/registerRoutes.js
@@ -4,9 +4,9 @@ import React from 'react';
 
 export const registerRoutes = function () {
     registry.add('adminRoute', 'manageGroups', {
-        targets: ['administration-server-usersAndRoles:1'],
+        targets: ['administration-server-usersAndRoles:20'],
         requiredPermission: 'adminGroups',
-        icon: <Group/>,
+        icon: null,
         label: 'siteSettings:groups.label',
         isSelectable: true,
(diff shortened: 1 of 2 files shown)


src/javascript/registrations/groups/registerRoutes.js
```diff
@@ -4,9 +4,9 @@ import React from 'react';
 
 export const registerRoutes = function () {
     registry.add('adminRoute', 'manageGroups', {
-        targets: ['administration-server-usersAndRoles:1'],
+        targets: ['administration-server-usersAndRoles:20'],
         requiredPermission: 'adminGroups',
-        icon: <Group/>,
+        icon: null,
```
AKarmanov: We need to remove Group from import as well, otherwise we can have problems building the module due to eslint issues

fpral: No, because this same file also registers a group site settings for Adminstration/Sites (look at line 18 of the file), where the icon is needed. But in Administration/Server, we shall not have an icon.

PR #88, opened 2023-04-26
TECH-1230 Fix test

Janin-Michel-Mathias:
## JIRA

https://jira.jahia.org/browse/TECH-1230

## Description
Fix test to avoid flakyness

### DIFF
--- a/tests/cypress/e2e/languageSettingsUI.cy.ts
+++ b/tests/cypress/e2e/languageSettingsUI.cy.ts
@@ -20,12 +20,17 @@ describe('Tests on UI for language settings', () => {
         LanguageSettings.visit('languagesTestSite', 'ca')
         languageSettings.switchReplaceUntranslatedWithDefault()
         languageSettings.submitChanges()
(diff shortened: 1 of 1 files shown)


tests/cypress/e2e/languageSettingsUI.cy.ts
```diff
@@ -20,6 +20,8 @@ describe('Tests on UI for language settings', () => {
         LanguageSettings.visit('languagesTestSite', 'ca')
         languageSettings.switchReplaceUntranslatedWithDefault()
         languageSettings.submitChanges()
+        //eslint-disable-next-line
+        cy.wait(5000)
```
jsinovassin, line 24: Please avoid to cy.wait. 
In your case, you should be able to use a waitUntil on the getNodeByPath
```
cy.waitUntil(() =>
            getNodeByPath('/sites/languagesTestSite', ['j:mixLanguage']).then((result) => result.data.jcr.nodeByPath.properties[0].value === true),
```

jsinovassin (approved):

PR #110, opened 2024-01-10
BACKLOG-22015 Remove html filtering UI

AKarmanov:
## JIRA

https://jira.jahia.org/browse/BACKLOG-22015 

## Description

Remove html filtering UI


gflores-jahia (commented):
I see we also have anthracite htmlfiltering reference here which I think we can also remove https://github.com/Jahia/siteSettings/blob/c76819e4f40a1e30e4529c416bb2f010e97f2d02/src/main/import/repository.xml#L243

gflores-jahia (approved):

PR #130, opened 2025-01-10
QA-15279: Show context menu for page models

baptistegrimaud:
https://jira.jahia.org/browse/QA-15279

## Description

Show context menu (right-click) for page models that was present in Jahia 7.x but is no longer present in Jahia 8.x.

### Details
The easy solution is to load the GWT page used in Jahia 7.x (with a functional context menu) as an Iframe and use the special parameters `redirect=false` and `fullscreen=true` to respectively prevent the redirect towards a Jcontent page and to display the page without the left navigation menu.
A cleaner (but more complex) solution would be to rewrite this page as a React page and use the `DisplayActions` React component.

### DIFF
--- a/src/javascript/registrations/jcontent/registerRoutes.js
+++ b/src/javascript/registrations/jcontent/registerRoutes.js
@@ -8,6 +8,6 @@ export const registerRoutes = function () {
         icon: <WebPage/>,
         label: 'siteSettings:models.label',
         isSelectable: true,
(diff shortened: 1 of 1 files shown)


baptistegrimaud:
@gflores-jahia , do you happen to know if we maintain a list of URLs/pages still relying on GWT that I could update with this PR? Thanks

jkevan (approved):

PR #158, opened 2025-07-29
feat: do not display inherited groups and users in user management.

dgriffon:
### Description
From now on, inherited groups and users are no more marked as belonging to a given group

When editing group members, now only direct members of a group are checked 

Screenshot 2025-07-29 at 18 59 30

In this example, if *groupC* is a member of *groupB* before the changes, both were checked. Now only the belonging group is checked. 

Information panel has been updated to inform about this new behavior. 
```
Add/Remove users as members of your group. Members of subgroups are not listed.
```
Groups and users are following the same rule. 

#### Source code

#### Tests

> [!TIP]

### DIFF
--- a/src/main/java/org/jahia/modules/sitesettings/groups/ManageGroupsFlowHandler.java
+++ b/src/main/java/org/jahia/modules/sitesettings/groups/ManageGroupsFlowHandler.java
@@ -458,9 +458,9 @@ public class ManageGroupsFlowHandler implements Serializable {
         });
 
         Set<JCRGroupNode> groups = PrincipalViewHelper.getGroupSearchResult(null, siteKey, null, null, null, null);
-
+        List<JCRNodeWrapper> members = groupNode.getMembers();
         for (JCRGroupNode group : groups) {
-            sortedResults.put(group, groupNode.getMembers().contains(group));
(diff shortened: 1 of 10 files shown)


jkevan (commented):
I am a bit concerned about performances.
From the past we encoutered terrible performances when displaying list of groups and users.
I know that the `.isMember()` method is optimized to be fast due to internal membership paths cached in memory. 
So replacing it by manual member checking may have an impact on those UIs depending on the number of users/groups/members.

How could we make sure we are not introducing a performance regression here ?

dgriffon:
> I am a bit concerned about performances. From the past we encoutered terrible performances when displaying list of groups and users. I know that the `.isMember()` method is optimized to be fast due to internal membership paths cached in memory. So replacing it by manual member checking may have an impact on those UIs depending on the number of users/groups/members.
> 
> How could we make sure we are not introducing a performance regression here ? I think it should be ok, since we are checking direct members and no recursive membership is done, but just wanna make sure. (I don't remember if our current perf tests contains checks on those UIs or not)

you're right, the code can be improved to not resolve the members in the loop.

jkevan (approved):

PR #171, opened 2025-12-31
fix: improve separator input

Janin-Michel-Mathias:
### Description
Improve csv separator input and add tests for bulk create users feature

#### Source code

#### Tests

> [!TIP]

### DIFF
--- a/src/main/java/org/jahia/modules/sitesettings/users/management/CsvFile.java
+++ b/src/main/java/org/jahia/modules/sitesettings/users/management/CsvFile.java
@@ -36,7 +36,9 @@ public class CsvFile implements Serializable {
     }
 
     public void setCsvSeparator(String csvSeparator) {
-        this.csvSeparator = csvSeparator;
+        if(csvSeparator.length() == 1) {
+            this.csvSeparator = csvSeparator;
+        }
     }
 
     public MultipartFile getCsvFile() {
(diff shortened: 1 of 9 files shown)


src/main/java/org/jahia/modules/sitesettings/users/management/CsvFile.java
```diff
@@ -36,7 +36,9 @@ public String getCsvSeparator() {
     }
 
     public void setCsvSeparator(String csvSeparator) {
-        this.csvSeparator = csvSeparator;
+        if(csvSeparator.length() == 1) {
```
jayblanc: Such silent assignation without exception or error is not good practice.

src/main/resources/jnt_siteSettingsManageUsers/html/siteSettingsManageUsers.flow/bulkCreateUser.jsp
```diff
@@ -55,7 +55,7 @@
                     <input type="file" name="csvFile" id="csvFile"/>
                 </div>
                 <label for="csvSeparator"><fmt:message key="label.csvSeparator"/></label>
-                <input class="span6" type="text" name="csvSeparator" value="${csvFile.csvSeparator}" id="csvSeparator"/>
+                <input class="span6" type="text" name="csvSeparator" value="${csvFile.csvSeparator}" id="csvSeparator" maxLength="1"/>
```
jayblanc: Even if the backend accept a char, keeping the constraint in the front is a good practice avoiding to handle backend error and preventing bad input early.

jayblanc (changes requested):

jayblanc (approved):

PR #190, opened 2026-04-07
refactor: New React UI for language site settings

tleclere-jahia:
### Description
New React UI for language site settings

#### Source code
- GraphQL custom endpoint
- React component (using states)

#### Tests
- Cypress tests for GraphQL
- Cypress tests for UI


src/main/java/org/jahia/modules/sitesettings/languages/GqlAdminQueryExtension.java
```diff
@@ -0,0 +1,31 @@
+package org.jahia.modules.sitesettings.languages;
+
+import graphql.annotations.annotationTypes.GraphQLDescription;
+import graphql.annotations.annotationTypes.GraphQLField;
+import graphql.annotations.annotationTypes.GraphQLName;
+import graphql.annotations.annotationTypes.GraphQLNonNull;
+import graphql.annotations.annotationTypes.GraphQLTypeExtension;
+import org.codehaus.plexus.util.StringUtils;
```
jkevan: Be careful of import and IDE autocompletion, this is a package coming from jahia -impl but will most likely be deprecated in Jahia 8.2.4.0.
Use common lang libs instead.

src/javascript/registrations/settings/LanguageSettings/LanguageModal.jsx
```diff
@@ -0,0 +1,224 @@
...
+                                      case 'inactive':
+                                          selectedLanguage.activeInEdit = false;
+                                          selectedLanguage.activeInLive = false;
+                                          selectedLanguage.mandatory = false;
+                                          break;
+                                      default:
+                                          break;
+                                  }
+
+                                  setSelectedLanguage({...selectedLanguage});
```
jkevan, line 160: You're mutating the current state object in-place, then spreading it into a "new" object. React may skip re-renders or produce stale closures since the object reference was already mutated before the state update. This should build a fresh object without touching the original.

src/javascript/registrations/settings/LanguageSettings/LanguageSettings.jsx
```diff
@@ -0,0 +1,130 @@
...
+        activeInEdit: false,
+        activeInLive: false,
+        mandatory: false
+    });
+    const [modalOpen, setModalOpen] = useState(false);
+
+    const allowsUnlistedLanguages = data?.jcr?.result?.site?.allowsUnlistedLanguages?.booleanValue;
+    const mixLanguage = data?.jcr?.result?.site?.mixLanguage?.booleanValue;
+    const defaultLanguage = data?.jcr?.result?.site?.defaultLanguage;
+    useMemo(() => setSiteLocales(data?.jcr?.result?.site?.siteLocales || []), [data]);
```
jkevan: Calling a state setter inside `useMemo` is a side effect — this should be `useEffect`.

src/javascript/registrations/settings/LanguageSettings/LanguageModal.jsx
```diff
@@ -0,0 +1,224 @@
...
+    site: PropTypes.string.isRequired,
+    selectedLanguage: PropTypes.string.isRequired,
+    setSelectedLanguage: PropTypes.func.isRequired,
+    isOpen: PropTypes.bool.isRequired,
+    closeModal: PropTypes.func.isRequired,
+    refetch: PropTypes.func.isRequired,
+    availableLocales: PropTypes.array.isRequired,
+    siteLocales: PropTypes.array.isRequired,
+    defaultLanguage: PropTypes.string.isRequired
+};
```
jkevan, line 203: `LanguageModal.propTypes` declares `selectedLanguage: PropTypes.string.isRequired` but the component receives an object with `{isNew, language, activeInEdit, activeInLive, mandatory, ...}`. Should be `PropTypes.shape({...})`.

src/javascript/registrations/settings/LanguageSettings/LanguageModal.jsx
```diff
@@ -0,0 +1,224 @@
...
+            }
+          }
+        }`);
+
+    const save = (l, addLanguage) => {
+        if (addLanguage) {
+            siteLocales = [...siteLocales, l];
+        } else {
+            siteLocales = siteLocales.map(lang => lang.language === l.language ? l : lang);
+        }
```
jkevan: The `save` function reassigns the `siteLocales` prop:

```js
const save = (l, addLanguage) => {
    if (addLanguage) {
        siteLocales = [...siteLocales, l];
    } else {
        siteLocales = siteLocales.map(lang => lang.language === l.language ? l : lang);
    }
```

This modifies the prop's binding within the function scope. Use a local variable instead:

```js
const save = (updatedLanguage, addLanguage) => {
    const updatedLocales = addLanguage
        ? [...siteLocales, updatedLanguage]
        : siteLocales.map(lang => lang.language === updatedLanguage.language ? updatedLanguage : lang);

    gqlSave({
        variables: {
            // ... use updatedLocales here
        }
    });
};
```

This also fixes the shadowed parameter `l` used in the inner `.filter()` / `.map()` callbacks.

src/javascript/registrations/settings/LanguageSettings/LanguageSettings.jsx
```diff
@@ -0,0 +1,130 @@
...
+
+    const openModal = language => {
+        language.isNew = !language.language;
+        setSelectedLanguage(language);
+        setModalOpen(true);
+    };
```
jkevan, line 47: This mutates the incoming `language` object (which comes from `siteLocales` state). Should create a new object:

```js
const openModal = language => {
    setSelectedLanguage({...language, isNew: !language.language});
    setModalOpen(true);
};
```

src/javascript/registrations/settings/LanguageSettings/LanguageContent.jsx
```diff
@@ -0,0 +1,218 @@
...
+                                                      } else {
+                                                          setDefaultLanguage(l.language);
+                                                      }
+                                                  }}/>
+                                        <MenuItem label={t('label.table.actions.delete.title')}
+                                                  iconStart={<Delete/>}
+                                                  isDisabled={l.mandatory || l.activeInEdit || l.language === defaultLanguage || l.language === uilang || languageCount > 0}
+                                                  onClick={() => {
+                                                      if (!l.mandatory && !l.activeInEdit && l.language !== defaultLanguage && l.language !== uilang && languageCount === 0) {
+                                                          deleteLanguage(l.language);
```
jkevan: The delete `MenuItem` computes `isDisabled` from a condition, and the `onClick` handler re-checks the exact same condition. The `onClick` branch that shows `alert()` should never fire if `isDisabled` works correctly. Pick one approach: either rely on `isDisabled` and simplify the `onClick`, or do the full validation in `onClick` and drop `isDisabled`. Having both is maintenance overhead.

src/javascript/registrations/settings/LanguageSettings/LanguageContent.jsx
```diff
@@ -0,0 +1,218 @@
...
+    const [anchorEl, setAnchorEl] = useState({});
+    const closeMenu = language => {
+        setMenuOpen(Object.assign({}, menuOpen, {[language]: false}));
+        setAnchorEl(Object.assign({}, anchorEl, {[language]: null}));
+    };
+
+    const openMenu = (e, language) => {
+        setMenuOpen(Object.assign({}, menuOpen, {[language]: !menuOpen[language]}));
+        setAnchorEl(Object.assign({}, anchorEl, {[language]: e.currentTarget}));
+    };
```
jkevan: The per-language `menuOpen` and `anchorEl` state using `Object.assign({}, ...)` is verbose. Since only one menu can be open at a time, simplify to:

```js
const [openMenuLang, setOpenMenuLang] = useState(null);
const [anchorEl, setAnchorEl] = useState(null);
```

This eliminates the stale-state risk and simplifies all open/close logic.

src/javascript/registrations/settings/LanguageSettings/LanguageSettings.jsx
```diff
@@ -0,0 +1,130 @@
...
+                               <LanguageContent site={site}
+                                                uilang={uilang}
+                                                openModal={openModal}
+                                                siteLocales={siteLocales}
+                                                defaultLanguage={defaultLanguage}
+                                                refetch={refetch}/>
+                               <UntranslatedContent site={site}
+                                                    value={allowsUnlistedLanguages && mixLanguage ? 'all' :
+                                                        !allowsUnlistedLanguages && mixLanguage ? 'only' :
+                                                            !allowsUnlistedLanguages && !mixLanguage ? 'never' : 'never'}
```
jkevan: This inline ternary chain deriving the untranslated content value is hard to scan.

This can be simplified — the last two branches both return `'never'`, so:

```js
const untranslatedValue = allowsUnlistedLanguages && mixLanguage
    ? 'all'
    : mixLanguage
        ? 'only'
        : 'never';
```

Or as a lookup if you prefer readability:

```js
const untranslatedValue = {
    'true-true': 'all',
    'false-true': 'only',
}[`${!!allowsUnlistedLanguages}-${!!mixLanguage}`] || 'never';
```

src/javascript/registrations/settings/LanguageSettings/UntranslatedContent.jsx
```diff
@@ -0,0 +1,123 @@
...
+        gqlSave({
+            variables: {
+                path: `/sites/${site}`,
+                mixLanguage,
+                allowsUnlistedLanguages
+            }
+        }).then(() => {
+            setModalOpen(false);
+            refetch();
+        });
```
jkevan: The `save` function initializes two variables, then overwrites them in every branch of a switch. A simple lookup object replaces the whole thing:

```js
const save = () => {
    const options = {
        never: {mixLanguage: false, allowsUnlistedLanguages: false},
        only:  {mixLanguage: true,  allowsUnlistedLanguages: false},
        all:   {mixLanguage: true,  allowsUnlistedLanguages: true},
    };
    gqlSave({
        variables: {path: `/sites/${site}`, ...options[untranslatedValue]}
    }).then(() => {
        setModalOpen(false);
        refetch();
    });
};
```

src/main/java/org/jahia/modules/sitesettings/languages/GqlLocale.java
```diff
@@ -0,0 +1,40 @@
...
+        super(locale.toString(), activeInEdit, activeInLive, mandatory);
+        this.locale = locale;
+        this.count = count;
+    }
+
+    @GraphQLField
+    @GraphQLDescription("Count locale usage")
+    public long getCount() {
+        return count;
+    }
```
jkevan, line 21: > Best strategy anyway is to move GQL backend from current project to standard core APIs. like suggested in global comment, so please don't spent to much effort addressing this comment in current architecture. (to reconsider/fix properly once moved to GraphQL core)

The `count` is calculated in the constructor via `countLanguages()`, meaning it executes a JCR SQL2 query for every locale regardless of whether the `count` field is requested in the GraphQL query. This defeats one of GraphQL's core benefits — clients only pay for what they ask for.

The fix is to make the computation lazy by moving it into the getter with simply deferring the calculation:

```java
@GraphQLField
@GraphQLDescription("Count locale usage")
public long getCount() {
    if (count == null) {
        count = countLanguages(this.language);
    }
    return count;
}
```

That's the `resolver pattern`, each field in a GraphQL schema has its own resolver function that only executes when the client's query actually requests that field. This is what makes GraphQL efficient by design: you only pay the cost of computing/fetching the fields the client asks for.

src/main/java/org/jahia/modules/sitesettings/languages/GqlJcrSiteExtension.java
```diff
@@ -0,0 +1,51 @@
...
+    private long countLanguages(String language) {
+        try {
+            return gqlJcrSite.getNode().getSession().getWorkspace().getQueryManager()
+                    .createQuery("SELECT count AS [rep:count(skipChecks=1)] FROM [jnt:translation]" +
+                                    " WHERE ISDESCENDANTNODE(['/sites/" + gqlJcrSite.getSiteKey() + "'])" +
+                                    " AND [jcr:language] = '" + JCRContentUtils.sqlEncode(language) + "'",
+                            Query.JCR_SQL2)
+                    .execute().getRows().nextRow().getValue("count").getLong();
+        } catch (RepositoryException e) {
+            return 0;
```
jkevan: Returning `0` on failure is indistinguishable from "this language has no content." A language with 500 pages of content could silently report 0 if the query fails, potentially leading an admin to delete it thinking it's unused. At minimum log the exception. But ideally, let it propagate so the GraphQL layer can return a partial error response, which is what GraphQL is designed for.

src/main/java/org/jahia/modules/sitesettings/languages/GqlAdminQueryExtension.java
```diff
@@ -0,0 +1,31 @@
...
+    private GqlAdminQueryExtension() {
+    }
+
+    @GraphQLField
+    @GraphQLDescription("List all available locales in the JVM")
+    public static Set<GqlLocale> getAvailableLocales(@GraphQLName("language") @GraphQLNonNull String language) {
+        return LanguageCodeConverters.getSortedLocaleList(LanguageCodeConverters.languageCodeToLocale(language)).stream()
+                .filter(l -> StringUtils.isNotBlank(l.toString()))
+                .map(GqlLocale::new)
+                .sorted(Comparator.comparing(l -> l.getDisplayName(language)))
```
jkevan: > Best strategy anyway is to move GQL backend from current project to standard core APIs. like suggested in global comment, so please don't spent to much effort addressing this comment in current architecture. (to reconsider/fix properly once moved to GraphQL core)

The server-side sort on `displayLanguage` introduces a redundant computation issue. `getAvailableLocales("en")` computes the display name for each locale during the query phase to perform the sort, and then each `GqlLocale.getDisplayName(displayLanguage)` computes it again when the field is resolved. The display name should only be computed once.

Two possible approaches:

1. **Remove the server-side sort and sort client-side** — drop the `language` parameter from the query phase entirely, let the frontend handle ordering (which it already does via `.sort((a, b) => a.displayName.localeCompare(b.displayName))` in `LanguageContent`).
2. **Cache the computed display name in the GQL object** — have the sort phase store the result so that `getDisplayName` returns the pre-computed value instead of recalculating.

src/main/java/org/jahia/modules/sitesettings/languages/GqlAdminQueryExtension.java
```diff
@@ -0,0 +1,31 @@
...
+
+@GraphQLTypeExtension(GqlAdminQuery.class)
+public class GqlAdminQueryExtension {
+    private GqlAdminQueryExtension() {
+    }
+
+    @GraphQLField
+    @GraphQLDescription("List all available locales in the JVM")
+    public static Set<GqlLocale> getAvailableLocales(@GraphQLName("language") @GraphQLNonNull String language) {
+        return LanguageCodeConverters.getSortedLocaleList(LanguageCodeConverters.languageCodeToLocale(language)).stream()
```
jkevan: I think we first need to transform `LanguageCodeConverters.languageCodeToLocale(language)` and check it's valid, not null, if not it mean language param is incorrect and an error should be throw in that case.

src/main/java/org/jahia/modules/sitesettings/languages/GqlAdminQueryExtension.java
```diff
@@ -0,0 +1,31 @@
...
+                .filter(l -> StringUtils.isNotBlank(l.toString()))
+                .map(GqlLocale::new)
```
jkevan: > Best strategy anyway is to move GQL backend from current project to standard core APIs. like suggested in global comment, so please don't spent to much effort addressing this comment in current architecture. (to reconsider/fix properly once moved to GraphQL core)

Using same object as site local is a bit strange, as a lot of data/field is not meaningful for detach locale (not linked to a given site.)

src/main/java/org/jahia/modules/sitesettings/languages/GqlJcrSiteExtension.java
```diff
@@ -0,0 +1,51 @@
...
+@GraphQLTypeExtension(GqlJcrSite.class)
+public class GqlJcrSiteExtension {
+    private final GqlJcrSite gqlJcrSite;
+
+    public GqlJcrSiteExtension(GqlJcrSite gqlJcrSite) {
+        this.gqlJcrSite = gqlJcrSite;
+    }
+
+    @GraphQLField
+    @GraphQLDescription("Count languages usage")
```
jkevan: Description looks incorrect

src/main/java/org/jahia/modules/sitesettings/languages/GqlLocale.java
```diff
@@ -0,0 +1,40 @@
+package org.jahia.modules.sitesettings.languages;
+
+import graphql.annotations.annotationTypes.GraphQLDescription;
+import graphql.annotations.annotationTypes.GraphQLField;
+import org.jahia.modules.graphql.provider.dxm.site.GqlSiteLanguage;
+
+import java.util.Locale;
+
+public class GqlLocale extends GqlSiteLanguage {
+    private final Locale locale;
```
jkevan: > Best strategy anyway is to move GQL backend from current project to standard core APIs. like suggested in global comment, so please don't spent to much effort addressing this comment in current architecture. (to reconsider/fix properly once moved to GraphQL core)

locale is only used for equals/hashcode I doubt that holding a ref to this object is very usefull, the String representation is probably lighter and enough the need.

src/javascript/registrations/settings/LanguageSettings/LanguageContent.jsx
```diff
@@ -0,0 +1,218 @@
...
+        const filteredSiteLocales = siteLocales.filter(l => l.language !== language);
+        gqlDeleteLanguage({
+            variables: {
+                path: `/sites/${site}`,
+                languages: filteredSiteLocales.filter(l => l.activeInEdit || l.activeInLive || l.defaultLanguage || l.mandatory).map(l => l.language),
+                mandatoryLanguages: filteredSiteLocales.filter(l => l.mandatory).map(l => l.language),
+                inactiveLanguages: filteredSiteLocales.filter(l => !l.activeInEdit).map(l => l.language),
+                inactiveLiveLanguages: filteredSiteLocales.filter(l => !l.activeInLive).map(l => l.language)
+            }
+        }).then(() => {
```
jkevan, line 70: The same pattern for building `languages`, `mandatoryLanguages`, `inactiveLanguages`, and `inactiveLiveLanguages` from a locale array is duplicated across `LanguageContent.deleteLanguage` and `LanguageModal.save`. Extract into a shared helper:

```js
const buildLanguageVariables = (path, defaultLanguage, locales) => ({
    path,
    defaultLanguage,
    languages: locales.filter(l => l.activeInEdit || l.activeInLive || l.defaultLanguage || l.mandatory).map(l => l.language),
    mandatoryLanguages: locales.filter(l => l.mandatory).map(l => l.language),
    inactiveLanguages: locales.filter(l => !l.activeInEdit).map(l => l.language),
    inactiveLiveLanguages: locales.filter(l => !l.activeInLive).map(l => l.language),
});
```

src/javascript/registrations/settings/LanguageSettings/LanguageModal.jsx
```diff
@@ -0,0 +1,227 @@
...
+                                  setSelectedLanguage({...selectedLanguage});
+                              }}/>
+                </div>
+            </ModalBody>
+            <ModalFooter>
+                <Button size="big"
+                        variant="ghost"
+                        label={t('label.actions.cancel')}
+                        data-sel-role="cancel"
+                        onClick={() => onClose()}/>
```
jkevan: The `onClick={() => method()}` form creates a new anonymous function on every render for no reason. You'd only need the wrapper if you were passing arguments (like `() => method(someValue)`) or needed to prevent event object forwarding — but here we doesn't use any parameters, so it's harmless either way.
It's a minor thing, but it's a pattern that shows up in a few places across the files. Worth cleaning up for consistency in all the places.

`onClick={() => onClose(l)}/>`
can be simplified into: onClick={onClose}/>
(also true for any other functional prop other than `onClick`)

src/javascript/registrations/settings/LanguageSettings/LanguageContent.jsx
```diff
@@ -0,0 +1,218 @@
...
+
+                        return (
+                            <TableRow key={l.language} onClick={e => e.stopPropagation()}>
+                                <TableBodyCell width={columnsWidth.default}>{defaultLanguage === l.language ?
+                                    <Star color="blue"/> : ''}
+                                </TableBodyCell>
+                                <TableBodyCell width={columnsWidth.languages}>
+                                    {l.displayName} <Pill label={l.language}/>
+                                </TableBodyCell>
+                                <TableBodyCell width={columnsWidth.availability}>{
```
jkevan: This nested ternary for resolving availability appears in both `LanguageContent` and `LanguageModal`, and is also difficult to read.

Extract this into a shared helper:

```js
const getAvailability = ({activeInEdit, activeInLive, mandatory}) => {
    if (activeInEdit && activeInLive && mandatory) return 'required';
    if (activeInEdit && activeInLive) return 'active';
    if (activeInEdit) return 'inactiveInLive';
    return 'inactive';
};
```

Then both components just call `getAvailability(l)`. The `LanguageContent` rendering becomes `t(\`label.availability.${getAvailability(l)}.title\`)`.

src/main/java/org/jahia/modules/sitesettings/languages/GqlAdminQueryExtension.java
jkevan: > Best strategy anyway is to move GQL backend from current project to standard core APIs. like suggested in global comment, so please don't spent to much effort addressing this comment in current architecture. (to reconsider/fix properly once moved to GraphQL core)

First transform language code to Locale and check result != null before continue, will make the method safer.
Correctly handle error and validation, method should crash if language is not correct. `@GraphQLNonNull` already protect from null value, but what if an invalid language is passed ?

src/main/java/org/jahia/modules/sitesettings/languages/GqlJcrSiteExtension.java
```diff
@@ -0,0 +1,51 @@
...
+    public Set<GqlLocale> getSiteLocales(@GraphQLName("language") @GraphQLNonNull String language) {
+        JCRSiteNode siteNode = (JCRSiteNode) gqlJcrSite.getNode();
+        return Stream.concat(
+                        siteNode.getLanguagesAsLocales().stream().map(locale -> new GqlLocale(locale, true, siteNode.getActiveLiveLanguages().contains(locale.toString()), siteNode.getMandatoryLanguages().contains(locale.toString()), countLanguages(locale.toString()))),
+                        siteNode.getInactiveLanguagesAsLocales().stream().map(locale -> new GqlLocale(locale, false, false, siteNode.getMandatoryLanguages().contains(locale.toString()), countLanguages(locale.toString()))))
+                .sorted(Comparator.comparing(l -> l.getDisplayName(language)))
+                .collect(Collectors.toCollection(LinkedHashSet::new));
+    }
```
jkevan: > Best strategy anyway is to move GQL backend from current project to standard core APIs. like suggested in global comment, so please don't spent to much effort addressing this comment in current architecture. (to reconsider/fix properly once moved to GraphQL core)

Waht is the advantage over the method that already exists in GqlJcrSite ?

```
@GraphQLField
    @GraphQLName("languages")
    @GraphQLDescription("Site languages")
    public Collection<GqlSiteLanguage> getLanguages() {
        List<GqlSiteLanguage> result = new ArrayList<>();

        for (String s : siteNode.getLanguages()) {
            result.add(new GqlSiteLanguage(s, true, siteNode.getActiveLiveLanguages().contains(s), siteNode.getMandatoryLanguages().contains(s)));
        }
        for (String s : siteNode.getInactiveLanguages()) {
            result.add(new GqlSiteLanguage(s, false, false, siteNode.getMandatoryLanguages().contains(s)));
        }

        return result;
    }
```

this looks redundant we should rely on existing method as much as possible.

jkevan (changes requested):
## Overall Review

Thanks for the work on this — the component decomposition is solid, with `LanguageContent`, `LanguageModal`, and `UntranslatedContent` each owning a clear responsibility. The feature is functional and the structure makes sense.

That said, there are a number of issues across the PR that need to be addressed before we can merge. I've left detailed inline comments on each one, but here's a summary of the recurring themes:

### State & Prop Mutation

The most systemic issue is direct mutation of state objects and props — it shows up in the modal's `onChange`, in `openModal`, and in `save`. React expects state to be treated as immutable; mutating it directly can lead to subtle bugs where re-renders don't trigger as expected. This needs to be fixed consistently across all components.

### Error Handling on Mutations

Every GraphQL mutation (`.then()` chain) is missing error handling. A failed call silently swallows the error, can leave the UI in an inconsistent state (e.g. menu still open), and gives the user zero feedback. At minimum add `.catch()` blocks, or use the `onError` option from `useMutation`.

### Separation of Concerns

All GraphQL queries and mutations are currently inlined inside UI components. Extracting them into dedicated files — along with helper methods responsible for building mutation variables — would significantly improve readability and make the business logic easier to test and maintain independently from the UI layer.

### Duplicated Logic

Several patterns are copy-pasted across components (availability label resolution, mutation variable building). These should be extracted into shared helpers to keep things DRY and reduce the risk of the copies drifting apart over time.

### General Guidance

Many of the inline comments point to standard React and GraphQL best practices (immutable state updates, proper `useEffect` vs `useMemo` usage, PropTypes accuracy, etc.). These are easy to miss when you're not working in this stack daily — I'd recommend leveraging agentic AI tooling (e.g. Copilot, Claude) during development as a way to catch these patterns early and get quick feedback on idiomatic React/GraphQL code.

### GraphQL API Backend Architecture Recommendation

The current GraphQL API layer in this project has some design issues worth addressing:

- **Duplicated logic**: Several methods replicate what already exists in GraphQL core without adding real value.
- **Scope extensions with poor implementation**: Some additions like `count` extend core capabilities, but the implementation is suboptimal — for example, `count` runs a JCR query eagerly in the constructor regardless of whether the field is requested in the GraphQL query, violating GraphQL's lazy resolution principle.
- **Scattered API surface**: Having pieces of what should be core API functionality spread across this extension module makes the overall architecture harder to reason about and maintain.

**Recommended approach:** Add the missing methods directly into GraphQL core (`graphql-dxm-provider`) using standard GraphQL best practices. This would:

- Eliminate the need for the duplicated/extended GQL objects in this project (e.g. `GqlLocale` would become obsolete and the entire GQL backend layer here could potentially be removed).
- Centralize language-related API capabilities where they belong.
- Avoid over-designed extensions and the questions that come with them.

This is feasible since both `siteSettings` and `graphql-dxm-provider` are packaged as core modules in the Jahia distribution. The dependency can be made explicit via:

```xml
<jahia-depends>graphql-dxm-provider=3.6.1</jahia-depends>
```

This ensures compatibility and guarantees the required version providing the new language panel APIs is present.

src/javascript/registrations/settings/LanguageSettings/LanguageContent.jsx
tleclere-jahia: It is a bug in moonstone: you can click on disabled menuItem

jkevan: Ok, I see, then could we move the condition in a single place at least to avoid replicating it in both method ?

tleclere-jahia: Done in the last commit

src/javascript/registrations/settings/LanguageSettings/LanguageSettings.jsx
```diff
@@ -0,0 +1,101 @@
...
+            path: `/sites/${site}`,
+            displayLanguage: uilang
+        },
+        skip: !site
+    });
+
+    if (error) {
+        console.error(error);
+        throw new Error(error.message);
+    }
```
jkevan: The throw happens before the useState/useEffect calls. This violates the Rules of Hooks (hooks must be called in the same order on every render). Move all useState/useEffect calls above any conditional return/throw.

src/javascript/registrations/settings/LanguageSettings/LanguageSettings.jsx
```diff
@@ -0,0 +1,101 @@
...
+        activeInEdit: false,
+        activeInLive: false,
+        mandatory: false
+    });
+    const [modalOpen, setModalOpen] = useState(false);
+
+    const allowsUnlistedLanguages = data?.jcr?.result?.site?.allowsUnlistedLanguages?.booleanValue;
+    const mixLanguage = data?.jcr?.result?.site?.mixLanguage?.booleanValue;
+    const defaultLanguage = data?.jcr?.result?.site?.defaultLanguage;
+    useEffect(() => setSiteLocales(data?.jcr?.result?.site?.languages || []), [data]);
```
jkevan: This is an anti-pattern. `siteLocales` is never modified locally — it just duplicates `data.jcr.result.site.languages`. Just derive it:
`const siteLocales = data?.jcr?.result?.site?.languages ?? [];`
Removes a state, a render cycle, and a stale-data window.

src/javascript/registrations/settings/LanguageSettings/LanguageContent.jsx
```diff
@@ -0,0 +1,182 @@
...
+                        <TableHeadCell width={columnsWidth.default}>{t('label.table.th.default')}</TableHeadCell>
+                        <TableHeadCell width={columnsWidth.languages}>{t('label.table.th.languages')}</TableHeadCell>
+                        <TableHeadCell
+                            width={columnsWidth.availability}
+                        >{t('label.table.th.availability')}
+                        </TableHeadCell>
+                        <TableHeadCell/>
+                    </TableRow>
+                </TableHead>
+                <TableBody>{siteLocales.sort((a, b) => a.displayName.localeCompare(b.displayName))
```
jkevan: Array.sort mutates an Apollo-owned array

Apollo freezes cache results in dev; in prod the array is the same reference held by the cache. Sorting it mutates the cache, which can produce non-deterministic UI ordering after refetches.
Fix — copy first, and memoize so the sort only runs when input changes:
```
import React, {useMemo, useState} from 'react';

// inside the component:
const sortedLocales = useMemo(
    () => [...siteLocales].sort((a, b) => a.displayName.localeCompare(b.displayName)),
    [siteLocales]
);

// in JSX:
<TableBody>
    {sortedLocales.map(l => {
        // ...existing code...
    })}
</TableBody>
```
What useMemo buys you here:
✅ Stable reference between renders → safe to pass to memoized children later.
✅ Sort skipped when re-rendering due to menu state, hover, etc.
⚠️ Don't rely on it for correctness — React may drop the cache. The copy ([...siteLocales]) is the correctness fix; useMemo is the perf fix.

src/javascript/registrations/settings/LanguageSettings/LanguageModal.jsx
```diff
@@ -0,0 +1,203 @@
...
+                                      case 'inactive':
+                                          selectedLanguage.activeInEdit = false;
+                                          selectedLanguage.activeInLive = false;
+                                          selectedLanguage.mandatory = false;
+                                          break;
+                                      default:
+                                          break;
+                                  }
+
+                                  setSelectedLanguage({...selectedLanguage});
```
jkevan: Direct mutation of `selectedLanguage` prop

You write into the parent's state object before the spread. With React's strict mode double-invocation, or if the parent passes the same object reference elsewhere, this corrupts state.

Fix — replace the switch with a lookup map and produce a fresh object:
```
// Module scope (outside the component):
const AVAILABILITY_FLAGS = {
    active:         {activeInEdit: true,  activeInLive: true,  mandatory: false},
    inactiveInLive: {activeInEdit: true,  activeInLive: false, mandatory: false},
    required:       {activeInEdit: true,  activeInLive: true,  mandatory: true},
    inactive:       {activeInEdit: false, activeInLive: false, mandatory: false}
};

// In the Dropdown onChange:
onChange={(e, v) => {
    const flags = AVAILABILITY_FLAGS[v.value];
    if (!flags) return;
    setSelectedLanguage({...selectedLanguage, ...flags});
}}
```
Bonus: drops ~20 lines and makes adding a new availability mode a one-line change.

src/javascript/registrations/settings/LanguageSettings/UntranslatedContent.jsx
```diff
@@ -0,0 +1,109 @@
...
+} from '@jahia/moonstone';
+import styles from './LanguageSettings.scss';
+import PropTypes from 'prop-types';
+
+export const UntranslatedContent = ({site, value, refetch}) => {
+    const {t} = useTranslation('siteSettings');
+
+    const [modalOpen, setModalOpen] = useState(false);
+
+    const [untranslatedValue, setUntranslatedValue] = useState(value);
```
jkevan, line 24: Local state desync in UntranslatedContent

Initial value is captured once. If value updates after a refetch(), the modal still shows the old radio choice.
Fix — reset local state when the modal opens (cheapest, no extra effect on every prop change):
```
const openModal = () => {
    setUntranslatedValue(value);
    setModalOpen(true);
};
const closeModal = () => setModalOpen(false);

// Button:
onClick={openModal}
// Cancel button:
onClick={closeModal}
```
Alternative if you prefer effect-based sync:
`useEffect(() => { setUntranslatedValue(value); }, [value]);`

src/javascript/registrations/settings/LanguageSettings/LanguageContent.jsx
```diff
@@ -0,0 +1,182 @@
...
+                                                          if (l.activeInEdit) {
+                                                              errorMessage = t('label.table.actions.delete.error.activeInEdit');
+                                                          }
+
+                                                          if (languageCount > 0) {
+                                                              errorMessage = t('label.table.actions.delete.error.language', {count: languageCount});
+                                                          }
+
+                                                          // eslint-disable-next-line no-alert
+                                                          alert(errorMessage);
```
jkevan: alert is not really user friendly, don't we have a moonstone component for such user feedback UI messaging ?

src/javascript/registrations/settings/LanguageSettings/LanguageModal.jsx
```diff
@@ -0,0 +1,203 @@
...
+            setSelectedLanguage({
+                isNew: true,
+                activeInEdit: false,
+                activeInLive: false,
+                mandatory: false
+            });
+            closeModal(updatedLanguage, addLanguage);
+            refetch();
+        });
+    };
```
jkevan: The current mutations rely on imperative .then() chains.
This has several issues:

- No error handling — a failed save fails silently and leaves the modal open with no feedback.
- loading state is unused — the Save/Delete buttons can be clicked multiple times during the request, causing duplicate submissions.
- Manual refetch() is prop-drilled through three components instead of being managed by Apollo.
- Side effects (state reset, close, refetch) are mixed into the trigger function, making the flow harder to follow.

Could we refactor to the standard Apollo pattern ? declare side effects via `onCompleted` / `onError` on the hook, consume loading in the UI, and let Apollo refresh the parent query via `refetchQueries`.
```
const [saveLanguages, {loading: isSaving}] = useMutation(LanguageGraphQL.gqlSaveSiteLanguages, {
    refetchQueries: [{
        query: LanguageGraphQL.gqlGetSiteLanguages,
        variables: {path: `/sites/${site}`, displayLanguage: uilang}
    }],
    awaitRefetchQueries: true,
    onCompleted: () => {
        setSelectedLanguage(EMPTY_LANGUAGE);
        closeModal();
    },
    onError: err => {
        console.error(err);
        notify(t('label.modal.error.save'), 'error');
    }
});

const save = (updatedLanguage, addLanguage) => {
    const updatedLocales = addLanguage
        ? [...siteLocales, updatedLanguage]
        : siteLocales.map(l => l.language === updatedLanguage.language ? updatedLanguage : l);

    saveLanguages({
        variables: LanguageHelper.buildLanguageVariables(`/sites/${site}`, updatedLocales)
    });
};

// In JSX:
<Button label={t('label.actions.save')} isDisabled={isSaving} onClick={() => save(selectedLanguage, false)}/>
```

Apply the same pattern to `gqlSetDefaultLanguage`, the delete mutation, and the untranslated-content mutation. Once done, the refetch prop can be removed from `LanguageContent`, `LanguageModal`, and `UntranslatedContent`.

src/javascript/registrations/settings/LanguageSettings/LanguageContent.jsx
```diff
@@ -0,0 +1,182 @@
...
+                                </TableBodyCell>
+                                <TableBodyCell width={columnsWidth.availability}>
+                                    {t(`label.availability.${LanguageHelper.getAvailability(l)}.title`)}
+                                </TableBodyCell>
+                                <TableBodyCell align="right">
+                                    <Button size="big"
+                                            variant="ghost"
+                                            icon={<MoreVert/>}
+                                            onClick={e => openMenu(e, l.language)}/>
+                                    <Menu isDisplayed={menuOpen === l.language}
```
jkevan: Single shared <Menu> instead of one per row

Problem — A <Menu> is rendered for every row, even closed. With 30 languages that's 30 popovers in the DOM tree.

Fix — Render one <Menu> outside the rows, populate items based on which row is open:

```
const [menuState, setMenuState] = useState({anchorEl: null, language: null});

const openMenu = (e, language) => setMenuState({anchorEl: e.currentTarget, language});
const closeMenu = () => setMenuState({anchorEl: null, language: null});

const currentRow = useMemo(
    () => sortedLocales.find(l => l.language === menuState.language),
    [sortedLocales, menuState.language]
);

// In the row:
<Button icon={<MoreVert/>} onClick={e => openMenu(e, l.language)} aria-label={t('label.actions.more')}/>

// Outside the rows, once:
{currentRow && (
    <Menu isDisplayed={Boolean(menuState.anchorEl)}
          anchorEl={menuState.anchorEl}
          onClose={closeMenu}
          anchorElOrigin={{vertical: 'bottom', horizontal: 'left'}}
          transformElOrigin={{vertical: 'top', horizontal: 'left'}}>
        <MenuItem label={t('label.actions.edit')} iconStart={<Edit/>}
                  onClick={() => { editLanguage(currentRow); closeMenu(); }}/>
        {/* ...existing code... */}
    </Menu>
)}
```

src/javascript/registrations/settings/LanguageSettings/LanguageModal.jsx
```diff
@@ -0,0 +1,203 @@
...
+}) => {
+    const {t} = useTranslation('siteSettings');
+
+    const onClose = () => {
+        setSelectedLanguage({
+            isNew: true,
+            activeInEdit: false,
+            activeInLive: false,
+            mandatory: false
+        });
```
jkevan: Extract the empty-language constant
Files: LanguageSettings.jsx, LanguageModal.jsx
Problem — The same literal is repeated 3 times:
`{isNew: true, activeInEdit: false, activeInLive: false, mandatory: false}`
Fix — Add in `LanguageHelper.js`:
```
export const EMPTY_LANGUAGE = Object.freeze({
    isNew: true,
    activeInEdit: false,
    activeInLive: false,
    mandatory: false
});
```
Then import and use everywhere. Object.freeze guarantees nobody mutates the shared default (relates to item 4).

src/javascript/registrations/settings/LanguageSettings/Language.graphql.jsx
```diff
@@ -0,0 +1,62 @@
...
+    jcr(workspace: EDIT) {
+        mutateNode(pathOrId: $path) {
+            defaultLanguage: mutateProperty(name: "j:defaultLanguage") {
+                setValue(value: $defaultLanguage, type: STRING)
+            }
+        }
+    }
+}`;
+
+export const gqlSave = gql`mutation deleteSiteLanguage($path: String!, $languages: [String!]!, $mandatoryLanguages: [String!]!, $inactiveLanguages: [String!]!, $inactiveLiveLanguages: [String!]!) {
```
jkevan: Rename the misleading gqlSave mutation

Problem — The GraphQL operation name is deleteSiteLanguage but it's used to add, edit and delete. The JS export is gqlSave. In LanguageContent.jsx it's even re-aliased as gqlDeleteLanguage.
Fix
```
export const gqlSaveSiteLanguages = gql`
    mutation saveSiteLanguages($path: String!, $languages: [String!]!, $mandatoryLanguages: [String!]!, $inactiveLanguages: [String!]!, $inactiveLiveLanguages: [String!]!) {
        jcr {
            mutateNode(pathOrId: $path) {
                // ...existing code...
            }
        }
    }
`;
```
Update both call sites to import `gqlSaveSiteLanguages` and remove the `gqlDeleteLanguage` alias.

src/javascript/registrations/settings/LanguageSettings/LanguageModal.jsx
```diff
@@ -0,0 +1,203 @@
...
+                                      label: t('label.availability.active.title'),
+                                      description: t('label.availability.active.description'),
+                                      value: 'active'
+                                  },
+                                  {
+                                      label: t('label.availability.required.title'),
+                                      description: t('label.availability.required.description'),
+                                      value: 'required'
+                                  }
+                              ]}
```
jkevan: Move static data outside the component

Problem — The 24-line availability data array and `columnsWidth` are recreated on every render. Translations are the only dynamic part.

Fix
```
// Module scope:
const AVAILABILITY_VALUES = ['inactive', 'inactiveInLive', 'active', 'required'];
const DEFAULT_DISABLED_VALUES = new Set(['inactive', 'inactiveInLive']);

// Inside the component:
const availabilityData = useMemo(
    () => AVAILABILITY_VALUES.map(value => ({
        value,
        label: t(`label.availability.${value}.title`),
        description: t(`label.availability.${value}.description`),
        isDisabled: DEFAULT_DISABLED_VALUES.has(value) && selectedLanguage.language === defaultLanguage
    })),
    [t, selectedLanguage.language, defaultLanguage]
);
```
Do the same for `columnsWidth` in LanguageContent.jsx — declare it at module scope, not inside the function body.

src/javascript/registrations/settings/LanguageSettings/LanguageContent.jsx
```diff
@@ -0,0 +1,182 @@
...
+
+    const editLanguage = language => {
+        if (language) {
+            closeMenu();
+        }
+
+        openModal(language);
+    };
+
+    const getLanguageCount = l => siteLocales.find(lang => lang.language === l.language)?.count || 0;
```
jkevan: getLanguageCount is redundant

`l` already comes from `siteLocales`. Replace every call with `l.count ?? 0` and delete the function.

src/javascript/registrations/settings/LanguageSettings/LanguageModal.jsx
```diff
@@ -0,0 +1,203 @@
...
+    const {t} = useTranslation('siteSettings');
+
+    const onClose = () => {
+        setSelectedLanguage({
+            isNew: true,
+            activeInEdit: false,
+            activeInLive: false,
+            mandatory: false
+        });
+        closeModal(null, false);
```
jkevan: Something looks wrong with this arguments, they are ignored silently, may be a rest of previous implem not cleaned ?

src/javascript/registrations/settings/LanguageSettings/LanguageModal.jsx
```diff
@@ -0,0 +1,203 @@
...
+        gqlSave({
+            variables: LanguageHelper.buildLanguageVariables(`/sites/${site}`, updatedLocales)
+        }).then(() => {
+            setSelectedLanguage({
+                isNew: true,
+                activeInEdit: false,
+                activeInLive: false,
+                mandatory: false
+            });
+            closeModal(updatedLanguage, addLanguage);
```
jkevan: Same: Something looks wrong with this arguments, they are ignored silently, may be a rest of previous implem not cleaned ?

src/javascript/registrations/settings/LanguageSettings/Language.graphql.jsx
jkevan, line 1: File extensions: .jsx files with no JSX
Files: LanguageHelper.jsx, Language.graphql.jsx

Per the project's lint config and conventions, files without JSX should be `.js`. The GraphQL queries file in particular should follow the repo's `*.gql-queries.js` naming convention.
Fix
```
git mv LanguageHelper.jsx LanguageHelper.js
git mv Language.graphql.jsx Language.gql-queries.js
```
Update imports — the bare ./LanguageHelper / ./Language.graphql paths keep working without extensions, so call sites usually don't change.

src/javascript/registrations/settings/LanguageSettings/LanguageModal.jsx
```diff
@@ -0,0 +1,203 @@
...
+        activeInLive: PropTypes.bool.isRequired,
+        mandatory: PropTypes.bool.isRequired
+    }).isRequired,
+    setSelectedLanguage: PropTypes.func.isRequired,
+    isOpen: PropTypes.bool.isRequired,
+    closeModal: PropTypes.func.isRequired,
+    refetch: PropTypes.func.isRequired,
+    availableLocales: PropTypes.array.isRequired,
+    siteLocales: PropTypes.array.isRequired,
+    defaultLanguage: PropTypes.string.isRequired
```
jkevan: Tighten PropTypes

Problem — `selectedLanguage.language` and `displayName` are marked isRequired, but the initial empty state doesn't have those keys → PropTypes warnings.

Fix
```
LanguageModal.propTypes = {
    site: PropTypes.string.isRequired,
    selectedLanguage: PropTypes.shape({
        isNew: PropTypes.bool.isRequired,
        language: PropTypes.string,       // optional until selected
        displayName: PropTypes.string,    // optional until selected
        activeInEdit: PropTypes.bool.isRequired,
        activeInLive: PropTypes.bool.isRequired,
        mandatory: PropTypes.bool.isRequired
    }).isRequired,
    setSelectedLanguage: PropTypes.func.isRequired,
    isOpen: PropTypes.bool.isRequired,
    closeModal: PropTypes.func.isRequired,
    refetch: PropTypes.func.isRequired,
    availableLocales: PropTypes.arrayOf(PropTypes.shape({
        language: PropTypes.string.isRequired,
        displayName: PropTypes.string.isRequired
    })).isRequired,
    siteLocales: PropTypes.arrayOf(PropTypes.shape({
        language: PropTypes.string.isRequired,
        displayName: PropTypes.string.isRequired,
        activeInEdit: PropTypes.bool,
        activeInLive: PropTypes.bool,
        mandatory: PropTypes.bool,
        count: PropTypes.number
    })).isRequired,
    defaultLanguage: PropTypes.string   // can be undefined while loading
};
```
Apply equivalent shapes in `LanguageContent.propTypes.`

src/main/resources/javascript/locales/en.json
```diff
@@ -40,5 +37,75 @@
...
+  "label": {
+    "title": "Languages",
+    "header": "Manage languages for \"{{ siteName }}\"",
+    "actions": {
+      "add": "Add",
+      "cancel": "Cancel",
+      "edit": "Edit",
+      "save": "Save"
+    },
+    "unstranslatedContent": {
```
jkevan: i18n typo: unstranslatedContent → untranslatedContent
Fix: do search and replace

src/main/java/org/jahia/modules/sitesettings/languages/GqlSiteLanguageExtension.java
```diff
@@ -0,0 +1,41 @@
...
+import org.jahia.osgi.BundleUtils;
+import org.jahia.services.content.JCRContentUtils;
+
+import javax.jcr.RepositoryException;
+import javax.jcr.query.Query;
+
+@GraphQLTypeExtension(GqlSiteLanguage.class)
+public class GqlSiteLanguageExtension {
+    private final GqlSiteLanguage gqlSiteLanguage;
+    private final Long count;
```
jkevan: Is it me or this count variable is never assigned appart `null` in constructor ? :D

jkevan (changes requested):

src/main/java/org/jahia/modules/sitesettings/languages/GqlSiteLanguageExtension.java
tleclere-jahia: My bad! Does this variable need to be define only once in the  GraphQLField "count" just below instead of executing the JCR Query?

src/javascript/registrations/settings/LanguageSettings/LanguageContent.jsx
tleclere-jahia: Would you prefer that I use a modal component here?

sdusolle (dismissed):

src/javascript/registrations/settings/LanguageSettings/LanguageContent.jsx
jkevan: Don't know I let Felix answering this UI/UX question

src/javascript/registrations/settings/LanguageSettings/LanguageContent.jsx
```diff
@@ -0,0 +1,181 @@
...
+        closeMenu();
+        setModalErrorDescription(description);
+    };
+
+    const sortedLocales = useMemo(
+        () => [...siteLocales].sort((a, b) => a.displayName.localeCompare(b.displayName)),
+        [siteLocales]
+    );
+
+    const currentRow = useMemo(
```
jkevan: `menuState.language` (local) and `sortedLocales` (Apollo cache) are updated independently. Two consequences:
- Stale anchor / undefined row — if `sortedLocales` refetches and drops the open row (concurrent edit, mutation refetch racing with setMenuState, etc.), `currentRow` becomes `undefined` while `menuState.anchorEl` still holds a detached DOM node. Any handler firing in that frame throws on currentRow.language, and focus is lost on close.
- Mount/unmount churn — `{currentRow && <Menu/>}` remounts the menu on every open/close, breaking transitions and recreating the portal each time. Moonstone's `<Menu>` expects to stay mounted and be toggled via isDisplayed.

Two small changes would be nice to address this:
- Store the `row snapshot` (not the `language` key) in `menuState`, so action handlers operate on the row the user actually clicked, independent of later refetches.
- Always render `<Menu>`; drive visibility with `isDisplayed`. Render `MenuItem` children only when a `row` is selected.

Something like:

```
// ...existing code...
const [menuState, setMenuState] = useState({anchorEl: null, row: null});
const openMenu = (e, row) => setMenuState({anchorEl: e.currentTarget, row});
const closeMenu = () => setMenuState(prev => ({...prev, anchorEl: null}));

// remove the currentRow useMemo; use menuState.row directly
const currentRow = menuState.row;
// ...existing code... (isDisabled, mutations, etc. unchanged)

<Menu isDisplayed={Boolean(menuState.anchorEl)}      anchorEl={menuState.anchorEl}      anchorElOrigin={{vertical: 'bottom', horizontal: 'left'}}      transformElOrigin={{vertical: 'top', horizontal: 'left'}}      onClose={closeMenu}>
    {currentRow && (
        <>
            {/* ...existing MenuItem children... */}
        </>
    )}
</Menu>

// ...existing code...
onClick={e => openMenu(e, l)}   // pass the row object, not l.language
```

src/javascript/registrations/settings/LanguageSettings/LanguageModal.jsx
```diff
@@ -0,0 +1,187 @@
...
+                              isDisabled={!selectedLanguage.isNew}
+                              value={selectedLanguage.language}
+                              data-sel-role="languages"
+                              data={availableLocales.map(l => {
+                                  return {
+                                      iconEnd: <Pill label={l.language.toUpperCase()}/>,
+                                      id: l.language,
+                                      label: l.displayName,
+                                      value: l.language,
+                                      isDisabled: siteLocales.find(lang => lang.language === l.language)
```
jkevan: This passes the matched object (truthy) or `undefined`. May be better to guarantee boolean value here by using `Boolean(siteLocales.find(...))` or `siteLocales.some(...)`.

src/javascript/registrations/settings/LanguageSettings/LanguageModal.jsx
```diff
@@ -0,0 +1,187 @@
...
+                </div>
+            </ModalBody>
+            <ModalFooter>
+                <Button size="big"
+                        variant="ghost"
+                        label={t('label.actions.cancel')}
+                        data-sel-role="cancel"
+                        onClick={onClose}/>
+                {selectedLanguage.isNew ?
+                    <Button size="big"
```
jkevan, line 149: The `Save` button is guarded by `isSavingSiteLanguages`, why not the `Add` button ? seems the users could double-submit. 
Also: should the button be disabled when `!selectedLanguage.language` (no language picked yet).

src/javascript/registrations/settings/LanguageSettings/LanguageModal.jsx
```diff
@@ -0,0 +1,187 @@
...
+    );
+
+    const [gqlSaveSiteLanguages, {loading: isSavingSiteLanguages}] = useMutation(LanguageGraphQL.gqlSaveSiteLanguages,
+        {
+            refetchQueries: [{
+                query: LanguageGraphQL.gqlGetSiteLanguages,
+                variables: {path: `/sites/${site}`, displayLanguage: uilang}
+            }],
+            awaitRefetchQueries: true,
+            onCompleted: () => onClose(),
```
jkevan: `onClose` calls `setSelectedLanguage(emptyLanguage)`. If the modal stayed open (e.g. for the onError path), the form would still hold the previous selection on next open. Today it works only because the modal closes. To avoid this risk, you could be explicit: reset state in `useEffect` on `isOpen` going false, instead of the mutation callback.

Something like:
```
 // Reset transient modal state when the modal transitions from open -> closed.
    const wasOpenRef = useRef(isOpen);
    useEffect(() => {
        if (wasOpenRef.current && !isOpen) {
            setSelectedLanguage(emptyLanguage);
            setModalErrorDescription(null);
        }

        wasOpenRef.current = isOpen;
    }, [isOpen, setSelectedLanguage]);
```

And, then, `onCompleted`, just close: `onCompleted: closeModal`

src/javascript/registrations/settings/LanguageSettings/LanguageHelper.js
```diff
@@ -0,0 +1,23 @@
+export const buildLanguageVariables = (path, locales) => ({
+    path,
+    languages: locales.filter(l => l.activeInEdit || l.activeInLive || l.defaultLanguage || l.mandatory).map(l => l.language),
```
jkevan: Are you sure about: `l.defaultLanguage` doesn't seems to exists in a language definition object  (That's probably why people love TypeScript :D) ?

src/javascript/registrations/settings/LanguageSettings/LanguageModal.jsx
```diff
@@ -0,0 +1,187 @@
...
+                setModalErrorDescription(t('label.modal.error.description'));
+            }
+        });
+
+    const save = (updatedLanguage, addLanguage) => {
+        const updatedLocales = addLanguage ?
+            [...siteLocales, updatedLanguage] :
+            siteLocales.map(l => l.language === updatedLanguage.language ? updatedLanguage : l);
+
+        gqlSaveSiteLanguages({variables: LanguageHelper.buildLanguageVariables(`/sites/${site}`, updatedLocales)});
```
jkevan, line 79: Minor: but I see a lot of: `/sites/${site}` repetition. I would be cleaner to do a React context init once in the root parent: LanguageSettings.jsx. It would allows: - to avoid those repetitions
- to do lither parameters from parents to children, access init once data using context instead of cascading those props.
those one along with: uilang are coming from redux, and never updated in the entire language app, so it make sense to move them to a global context instead of each component signature. 
(But this is minor, of to not do it if you don't want to.)

src/javascript/registrations/settings/LanguageSettings/LanguageSettings.jsx
```diff
@@ -0,0 +1,99 @@
...
+import {useNodeChecks} from '@jahia/data-helper';
+import {LanguageContent} from './LanguageContent';
+import {UntranslatedContent} from './UntranslatedContent';
+import {LanguageModal} from './LanguageModal';
+import * as LanguageGraphQL from './Language.gql-queries';
+
+export const LanguageSettings = () => {
+    const {t} = useTranslation('siteSettings');
+    const {site, uilang} = useSelector(state => ({site: state.site, uilang: state.uilang}), shallowEqual);
+    const res = useNodeChecks({path: `/sites/${site}`, uilang}, {requiredPermission: 'siteAdminLanguages'});
```
jkevan, line 17: Minor: If the redux `site` is empty, the path becomes `/sites/undefined`, which may have weird side effects, may be we could leave the code earlier?

src/javascript/registrations/settings/LanguageSettings/LanguageSettings.jsx
```diff
@@ -0,0 +1,99 @@
...
+        setModalOpen(true);
+    };
+
+    const closeModal = () => {
+        setModalOpen(false);
+    };
+
+    if (error) {
+        console.error(error);
+        throw new Error(error.message);
```
jkevan: Sorry for not pointing it earlier, but I think it would be better not to throw Exception/Error like this. It may risk to triggers the React error boundary on every Apollo error (network blip, expired session, …). The best would be to returning a user facing error message and keep the user able to retry. But I see other place like this in other project files (SiteProperties), so for now it's minor, not perfect, but at least consistent with one other case.
(if fixing it, we should also do the same fix for SiteProperties. But it's ok for now not changing this and keeping this.)

src/javascript/registrations/settings/LanguageSettings/LanguageSettings.jsx
```diff
@@ -0,0 +1,99 @@
...
+        },
+        skip: !site
+    });
+
+    const [selectedLanguage, setSelectedLanguage] = useState({
+        isNew: true,
+        activeInEdit: false,
+        activeInLive: false,
+        mandatory: false
+    });
```
jkevan: Is this an empty Language ? like this one ? 

```
const emptyLanguage = Object.freeze({
    isNew: true,
    activeInEdit: false,
    activeInLive: false,
    mandatory: false
});
```
(defined in `LanguageModal.jsx` ?)

Would be nice to use it, could be moved and export by `LanguageHelper.js` for sharing.

src/javascript/registrations/settings/LanguageSettings/LanguageModal.jsx
```diff
@@ -0,0 +1,187 @@
...
+    activeInLive: false,
+    mandatory: false
+});
+const availabilityFlags = {
+    active: {activeInEdit: true, activeInLive: true, mandatory: false},
+    inactiveInLive: {activeInEdit: true, activeInLive: false, mandatory: false},
+    required: {activeInEdit: true, activeInLive: true, mandatory: true},
+    inactive: {activeInEdit: false, activeInLive: false, mandatory: false}
+};
+const availabilityValues = ['inactive', 'inactiveInLive', 'active', 'required'];
```
jkevan: Minor:

`availabilityValues` defines display order using array, but `availabilityFlags` is keyed differently. Combine into a single source of truth could be nice ?

Something like:
```
const AVAILABILITIES = [
    { value: 'inactive', flags: {...} },
    { value: 'inactiveInLive', flags: {...} },
    { value: 'active', flags: {...} },
    { value: 'required', flags: {...} }
];
```

src/javascript/registrations/settings/LanguageSettings/LanguageModal.jsx
```diff
@@ -0,0 +1,187 @@
...
+                                    description={modalErrorDescription}/>
+                <div className={styles.field}>
+                    <Typography variant="subheading">{t('label.modal.language.title')}</Typography>
+                    <Dropdown className={styles.dropdown}
+                              placeholder={t('label.modal.language.placeholder')}
+                              variant="outlined"
+                              isDisabled={!selectedLanguage.isNew}
+                              value={selectedLanguage.language}
+                              data-sel-role="languages"
+                              data={availableLocales.map(l => {
```
jkevan: Minor: `availableLocales.map(...)` recreates a new array each render → moonstone re-renders the dropdown content. Would it be interesting to consider a `useMemo` for this ?

src/javascript/registrations/settings/LanguageSettings/LanguageSettings.jsx
```diff
@@ -0,0 +1,99 @@
...
+        throw new Error(error.message);
+    }
+
+    if (!res.checksResult) {
+        return '';
+    }
+
+    return (
+        <LayoutContent isLoading={loading}
+                       aria-labelledby="language-settings"
```
jkevan: minor: `aria-labelledby="language-settings"` but no element has `id="language-settings"`.

src/javascript/registrations/settings/LanguageSettings/LanguageContent.jsx
```diff
@@ -0,0 +1,181 @@
...
+                                {l.displayName} <Pill label={l.language}/>
+                            </TableBodyCell>
+                            <TableBodyCell width={columnsWidth.availability}>
+                                {t(`label.availability.${LanguageHelper.getAvailability(l)}.title`)}
+                            </TableBodyCell>
+                            <TableBodyCell align="right">
+                                <Button size="big"
+                                        variant="ghost"
+                                        icon={<MoreVert/>}
+                                        aria-label={t('label.actions.more')}
```
jkevan: Minor: The `<Button icon={<MoreVert/>}>` per row has `aria-label={t('label.actions.more')}` — good. But all rows share the same label; Could be nice to include the language name for screen readers: `t('label.actions.moreFor', { name: l.displayName })`.

src/javascript/registrations/settings/LanguageSettings/LanguageSettings.jsx
```diff
@@ -0,0 +1,99 @@
...
+    const {t} = useTranslation('siteSettings');
+    const {site, uilang} = useSelector(state => ({site: state.site, uilang: state.uilang}), shallowEqual);
+    const res = useNodeChecks({path: `/sites/${site}`, uilang}, {requiredPermission: 'siteAdminLanguages'});
+
+    const {data, loading, error} = useQuery(LanguageGraphQL.gqlGetSiteLanguages, {
+        variables: {
+            path: `/sites/${site}`,
+            displayLanguage: uilang
+        },
+        skip: !site
```
jkevan, line 24: When skipped, data is `undefined` and `loading` is false. The `LayoutContent` `isLoading={loading}` then shows content with no data. This also refer to a previous comment that mention: may be it could be nice to leave earlier if no site to avoid side effect and this kind of check.

src/javascript/registrations/settings/LanguageSettings/LanguageHelper.js
```diff
@@ -0,0 +1,23 @@
+export const buildLanguageVariables = (path, locales) => ({
```
jkevan, line 7: `buildLanguageVariables` returns a fresh object — fine, but it'd be clearer as a pure named export with JSDoc.

Also `LanguageHelper.js` has no PropTypes/JSDoc; one line of doc on `getAvailability` describing the 4 states would help future maintainers.

src/javascript/registrations/settings/LanguageSettings/LanguageModal.jsx
```diff
@@ -0,0 +1,187 @@
...
+
+                <div className={styles.field}>
+                    <Typography variant="subheading">{t('label.modal.availability.title')}</Typography>
+                    <Dropdown data-sel-role="availability"
+                              data={availabilityData}
+                              placeholder={t('label.modal.availability.placeholder')}
+                              variant="outlined"
+                              className={styles.dropdown}
+                              data-value={LanguageHelper.getAvailability(selectedLanguage)}
+                              value={LanguageHelper.getAvailability(selectedLanguage)}
```
jkevan: We should avoid this kind of double call, duplicates.

src/main/java/org/jahia/modules/sitesettings/languages/GqlSiteLanguageExtension.java
jkevan: It would have been interesting as a cache for current thread/request.
in case count() would be called multiple time on the same Field, but this doesn't really make sense, I see you already removed it, good choice :)

jkevan (changes requested):

jkevan (dismissed):
Thanks a lot for all the fixes, it looks good to me !
Well done 👏

sdusolle (dismissed):

jkevan (dismissed):

PR #210, opened 2026-06-12
test: fix tests on new languages screen

sdusolle:
### Description
siteSettings-8.12.0-SNAPSHOT build is skipped because the tests are failing : https://github.com/Jahia/siteSettings/actions/runs/27360968206/job/80994183346
- refactored `Should be able to deactivate a non-mandatory language`
- check for h1 in `Navigates to edit languages`
- fixed workflow

#### Source code

#### Tests

> [!TIP]

### DIFF
--- a/tests/cypress/page-object/siteSettingsLanguages.ts
+++ b/tests/cypress/page-object/siteSettingsLanguages.ts
@@ -18,8 +18,8 @@ export class SiteSettingsLanguages extends BasePage {
         getComponent(Menu).select('Edit')
     }
 
-    checkLanguageAvailability(expectedLabel: string) {
-        getComponentByRole(Dropdown, 'availability').get().should('contain.text', expectedLabel)
+    checkLanguageAvailability(expectedValue: string) {
+        getComponentByRole(Dropdown, 'availability').get().should('have.attr', 'data-value', expectedValue)
     }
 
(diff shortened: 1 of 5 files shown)


tests/cypress/page-object/siteSettingsLanguages.ts
```diff
@@ -16,11 +16,10 @@ export class SiteSettingsLanguages extends BasePage {
...
         getComponent(Menu).select('Edit')
-        return getComponentByRole(Dropdown, 'availability')
-            .get()
-            .then(($el) => {
-                return $el.attr('data-value') === expectedValue
-            })
+    }
+
+    checkLanguageAvailability(expectedLabel: string) {
+        getComponentByRole(Dropdown, 'availability').get().should('contain.text', expectedLabel)
```
gflores-jahia: Any reason why we switched the check using attr vs contain.text ? Maybe backwards test compat ?

I feel targeted value is more preferable than a loose string matching (susceptible to break when the label change). Could be rewritten preferably like

```
getComponentByRole(Dropdown, 'availability').get()
  .should('have.attr', 'data-value', expectedValue)
```

but if it's for test backwards compatibility (when data-value is introduced) then ok to keep.

gflores-jahia (approved):

PR #222, opened 2026-07-08
fix: consistently escape special characters in page model names (Page Models admin)

jkevan:
### Summary
Page model **names and titles** containing special characters (`<`, `>`, `&`, quotes) were not consistently escaped when rendered in the **Page Models** administration screen, so such values could be displayed incorrectly.

### Change
HTML-escape `pageModel.displayableName` and `j:pageTemplateTitle` (`fn:escapeXml` / `c:out`) in the Page Models admin JSP (default + Bootstrap3/Material variants), mirroring the file's existing convention for the sibling site name. Fixed at the JSP layer: the shared core getter is intentionally raw and already escaped by its callers, so changing the getter would double-escape other call sites.

### Verification
Deploy-tested on a local 8.2.x instance: page model names and titles containing special characters now render correctly.

### DIFF
--- a/.chachalog/b3Xq9wRt.md
+++ b/.chachalog/b3Xq9wRt.md
@@ -1,0 +1,6 @@
+---
+# Allowed version bumps: patch, minor, major
+siteSettings: patch
+---
+
+Fixed display of page model names and titles containing special characters in the Page Models administration screen.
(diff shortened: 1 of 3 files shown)


jkevan:
Superseded by the rebased PR.

PR #239, opened 2026-07-27
fix: consistently escape special characters in Manage Users flow messages

eallais:
## Summary

The Manage Users settings flows report user names and display names back to the operator through
Spring WebFlow messages, and the flow views render those messages as HTML. Those reported values are
free text, and they now go through the same encoder the rest of the screen already applies to them —
`UserProperties.populate()` for the confirmation view-state, `fn:escapeXml` for the list-view
columns. A name holding markup characters (`<`, `>`, `&`, quotes) is shown as the characters it
contains.

## Change

Escape the free-text values where they enter a message, keeping the raw value for lookups and
persistence:

- `UsersFlowHandler.removeUser()` — the resolved display name, and the user key reported when the
  path does not resolve. This also covers `bulkDeleteUser()`, which delegates to it.
- `UsersFlowHandler.bulkAddUser()` — the name read from the CSV column, which is reported back in
  the "already exists" / "skipped" / "creation failed" messages.

Escaping is applied to the message argument rather than to `${message.text}` in the views on
purpose: several messages in this bundle intentionally carry markup (`<strong>{0}</strong>`, `<br/>`,
`<ul><li>`), so escaping the rendered message would display that markup literally. Handling it in the
flow handler also covers both settings skins at once, instead of duplicating the change across the
`.flow` variants.

The encoder is the one this module already uses for the same values (`StringEscapeUtils.escapeXml`,
commons-lang 2.x, as in `UserProperties.populate()`), so no double-escaping is introduced — the
display name returned by `PrincipalViewHelper.getDisplayName()` is raw at this point.

## Verification

- `mvn compile` passes.
- Extends the existing `manageUsersDisplayEscaping` Cypress regression with two cases that drive the
  real flows end to end — row action → confirmation view-state → confirm event for the single
  removal, and its own transition for the bulk removal — on users whose display name contains
  markup, then assert the resulting confirmation message shows that name as literal text: no live
  element carrying a handler, and the handler never runs.

### DIFF
--- a/src/main/java/org/jahia/modules/sitesettings/users/UsersFlowHandler.java
+++ b/src/main/java/org/jahia/modules/sitesettings/users/UsersFlowHandler.java
@@ -17,6 +17,7 @@ package org.jahia.modules.sitesettings.users;
 
 import au.com.bytecode.opencsv.CSVReader;
 import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang.StringEscapeUtils;
 import org.apache.commons.lang.StringUtils;
 import org.jahia.data.viewhelper.principal.PrincipalViewHelper;
 import org.jahia.modules.sitesettings.users.management.CsvFile;
@@ -146,10 +147,13 @@ public class UsersFlowHandler implements Serializable {
                         List<String> lineElementList = Arrays.asList(lineElements);
                         Properties properties = buildProperties(headerElementList, lineElementList);
                         String userName = lineElementList.get(userNamePos);
+                        // the name is reported back in HTML-rendered messages before it is known to
+                        // satisfy the name syntax: escape the reported form, keep the raw one for lookups
+                        String userNameText = StringEscapeUtils.escapeXml(userName);
                         String password = lineElementList.get(passwordPos);
                         if (userManagerService.userExists(userName, siteKey)) {
                              context.addMessage(new MessageBuilder().error().code(
-                                   "siteSettings.users.bulk.errors.user.already.exists").arg(userName).build());
+                                   "siteSettings.users.bulk.errors.user.already.exists").arg(userNameText).build());
                              hasErrors = true;
                         } else if (userManagerService.isUsernameSyntaxCorrect(userName)) {
                             PolicyEnforcementResult evalResult = pwdPolicyService.enforcePolicyOnUserCreate(userName, password);
@@ -157,10 +161,10 @@ public class UsersFlowHandler implements Serializable {
                                 JCRUserNode jahiaUser = userManagerService.createUser(userName, siteKey, password, properties, session);
                                 if (jahiaUser != null) {
                                     context.addMessage(new MessageBuilder().info().code(
-                                            "siteSettings.users.bulk.user.creation.successful").arg(userName).build());
+                                            "siteSettings.users.bulk.user.creation.successful").arg(userNameText).build());
                                 } else {
                                     context.addMessage(new MessageBuilder().error().code(
-                                            "siteSettings.users.bulk.errors.user.creation.failed").arg(userName).build());
+                                            "siteSettings.users.bulk.errors.user.creation.failed").arg(userNameText).build());
                                     hasErrors = true;
                                 }
                             } else {
@@ -170,12 +174,12 @@ public class UsersFlowHandler implements Serializable {
                                 }
                                 result.append("</ul>");
                                 context.addMessage(new MessageBuilder().error().code(
-                                        "siteSettings.users.bulk.errors.user.skipped.password").args(new String[]{userName, result.toString()}).build());
+                                        "siteSettings.users.bulk.errors.user.skipped.password").args(new String[]{userNameText, result.toString()}).build());
                                 hasErrors = true;
                             }
                         } else {
                             context.addMessage(new MessageBuilder().error().code(
-                                    "siteSettings.users.bulk.errors.user.skipped").arg(userName).build());
+                                    "siteSettings.users.bulk.errors.user.skipped").arg(userNameText).build());
                             hasErrors = true;
                         }
                     }
@@ -235,10 +239,12 @@ public class UsersFlowHandler implements Serializable {
                 JCRUserNode jahiaUser = userManagerService.lookupUserByPath(userKey);
                 if (jahiaUser == null || !isInAdministeredScope(jahiaUser.getPath())) {
                     context.addMessage(new MessageBuilder().error().code(
-                            "siteSettings.user.remove.unsuccessful").arg(userKey).build());
+                            "siteSettings.user.remove.unsuccessful").arg(StringEscapeUtils.escapeXml(userKey)).build());
                     return false;
                 }
-                String displayName = PrincipalViewHelper.getDisplayName(jahiaUser);
+                // messages are rendered as HTML by the flow views: escape the free-text display name,
+                // consistently with UserProperties.populate()
+                String displayName = StringEscapeUtils.escapeXml(PrincipalViewHelper.getDisplayName(jahiaUser));
                 if (userManagerService.deleteUser(jahiaUser.getPath(), session)) {
                     context.addMessage(new MessageBuilder().info().code(
                             "siteSettings.user.remove.successful").arg(displayName).build());
--- a/src/main/java/org/jahia/modules/sitesettings/users/UsersFlowHandler.java
+++ b/src/main/java/org/jahia/modules/sitesettings/users/UsersFlowHandler.java
@@ -147,8 +147,7 @@ public class UsersFlowHandler implements Serializable {
                         List<String> lineElementList = Arrays.asList(lineElements);
                         Properties properties = buildProperties(headerElementList, lineElementList);
                         String userName = lineElementList.get(userNamePos);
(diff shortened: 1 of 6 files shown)


src/main/java/org/jahia/modules/sitesettings/users/UsersFlowHandler.java
```diff
@@ -146,21 +147,24 @@ public Boolean doInJCR(JCRSessionWrapper session) throws RepositoryException {
                         List<String> lineElementList = Arrays.asList(lineElements);
                         Properties properties = buildProperties(headerElementList, lineElementList);
                         String userName = lineElementList.get(userNamePos);
+                        // the name is reported back in HTML-rendered messages before it is known to
+                        // satisfy the name syntax: escape the reported form, keep the raw one for lookups
+                        String userNameText = StringEscapeUtils.escapeXml(userName);
```
eallais, line 151: `userNameText` feeds four messages in `bulkAddUser()`, and the existing `manageUsersDisplayEscaping` spec only drives the removal path — so none of the four is exercised.

**Worth adding a small CSV upload case — one row whose `j:nodename` contains markup — asserting the resulting message renders it as text. Or, if that is more than this change should carry, say so and leave it; I'd rather it be a deliberate call than an oversight.**

src/main/java/org/jahia/modules/sitesettings/users/UsersFlowHandler.java
```diff
@@ -235,10 +239,12 @@ public Boolean doInJCR(JCRSessionWrapper session) throws RepositoryException {
...
                 if (jahiaUser == null || !isInAdministeredScope(jahiaUser.getPath())) {
                     context.addMessage(new MessageBuilder().error().code(
-                            "siteSettings.user.remove.unsuccessful").arg(userKey).build());
+                            "siteSettings.user.remove.unsuccessful").arg(StringEscapeUtils.escapeXml(userKey)).build());
                     return false;
                 }
-                String displayName = PrincipalViewHelper.getDisplayName(jahiaUser);
+                // messages are rendered as HTML by the flow views: escape the free-text display name,
+                // consistently with UserProperties.populate()
+                String displayName = StringEscapeUtils.escapeXml(PrincipalViewHelper.getDisplayName(jahiaUser));
```
eallais, line 245: This line covers the bulk-delete path too, but only because `bulkDeleteUser()` delegates to `removeUser()` — and no test holds that shape in place.

`bulkDeleteUser()` is a separate flow transition (`doUsersAction('bulkDeleteUser')`) reached by its own button, so a future refactor that builds its own message instead of delegating would quietly lose this handling, with every existing test still green.

**Either add a bulk-delete case to the spec, or drop a short comment on `bulkDeleteUser()` noting that its escaping lives in `removeUser()` — enough that the next person refactoring it knows what they'd be breaking.**

tests/cypress/e2e/manageUsersDisplayEscaping.test.cy.ts
```diff
@@ -58,4 +71,32 @@ describe('Manage Users - display name rendering', () => {
...
+        })
+
+        cy.get('#searchString').clear()
+        cy.get('#searchString').type(REMOVED_USER)
+        cy.get('[name="_eventId_search"]').first().click()
+
+        // drive the real removal: the row action leads to the confirmation view-state, whose confirm
+        // button posts the delete. force: true because one settings skin nests it in a modal.
+        cy.get('a[href="#delete"]', { timeout: 10000 }).first().click()
+        cy.get('[name="_eventId_confirm"]', { timeout: 10000 }).first().click({ force: true })
```
eallais, line 100: `force: true` is doing real work here — in the bootstrap3 skin this button lives inside a hidden `#confirmDeleteModal` — but it also means the test would still pass if the modal stopped opening entirely, since the button submits the form either way.

- That is acceptable for a rendering assertion, which is what this test is for. Just flagging that it is not a check of the confirmation UX.
- If you'd rather keep the user-visible path honest, clicking `[data-target="#confirmDeleteModal"]` first and dropping the force would do it — at the cost of a skin-specific selector the plain `.flow` variant doesn't have.

**Source:** `src/main/resources/jnt_siteSettingsManageUsers/html/siteSettingsManageUsers.settingsBootstrap3GoogleMaterialStyle.flow/removeUser.jsp` (confirm button inside `#confirmDeleteModal`) vs `.../siteSettingsManageUsers.flow/removeUser.jsp:118` (same button, not in a modal).

src/main/java/org/jahia/modules/sitesettings/users/UsersFlowHandler.java
eallais: Not covered, and that is a deliberate call. A CSV case was written for this and then withdrawn: it could not reliably observe the rendered message through the Manage-Users admin frame — the assertion never found the flash message, across two attempts. A red test is worse than a stated gap, so it is recorded on our side rather than dropped silently.

src/main/java/org/jahia/modules/sitesettings/users/UsersFlowHandler.java
eallais: Covered now — added a bulk-removal case that drives that transition end to end (its own view-state and confirm event) and asserts the same three things. A refactor that stops delegating goes red there. Green in CI.

tests/cypress/e2e/manageUsersDisplayEscaping.test.cy.ts
eallais: Acknowledged, keeping it as is. The assertion here is about what gets rendered, not about the confirmation UX, and the plain skin has no modal to open — a skin-specific selector would buy little.

src/main/java/org/jahia/modules/sitesettings/users/UsersFlowHandler.java
```diff
@@ -235,10 +239,12 @@ public Boolean doInJCR(JCRSessionWrapper session) throws RepositoryException {
                 JCRUserNode jahiaUser = userManagerService.lookupUserByPath(userKey);
                 if (jahiaUser == null || !isInAdministeredScope(jahiaUser.getPath())) {
                     context.addMessage(new MessageBuilder().error().code(
-                            "siteSettings.user.remove.unsuccessful").arg(userKey).build());
+                            "siteSettings.user.remove.unsuccessful").arg(StringEscapeUtils.escapeXml(userKey)).build());
```
jayblanc, line 241: Why are we filtering userKey ? isn't userKey supposed to be a path ?
If yes, and if it is direct user input, I think it is better to validate the entry as a valid path (so without xml) before using it on the lookupUserByPath() and thus it won't be required to filter it in the message.
This would also avoid any injection.

eallais: Right on the first half — `userKey` is direct input on one path: `bulkDeleteUser()` iterates `requestParameters.userToDelete` and calls `removeUser()` with each (`UsersFlowHandler:389`).

Two things against the suggestion though:

- `lookupUserByPath()` resolves a path, it does not build a query — `session.getNode(userKey)` (`JahiaUserManagerService:190-203`), every failure caught, malformed input just returns `null`.
- The encoding in `5f09a6d` is needed regardless: the same message key is emitted again at :255 with `displayName` (the user's `jcr:title`, free text). Validating `userKey` would not cover that one.

Agreed the path is better validated than relied on at the sink — the two are complementary. I would rather do it separately: it changes what happens on an unresolvable path, which is beyond this rendering fix, and wants its own test. Follow-up if you agree?

jayblanc (dismissed):

eallais:
Follow-up left out of this PR, recorded here while the context is fresh — no ticket opened yet.

**Hardening the `userKey` resolution** (from @jayblanc's point above). Two ways, neither obviously better:

- **Validate the path shape** before `lookupUserByPath()`. Matches the suggestion literally, but the regex would restate a rule that already lives in `userManagementUserNamePattern` (`jahia.properties`, default `[0-9a-z_A-Z\-\{\}\.@]+`). A deployment that widens that setting would get legitimate users rejected on removal, with nothing to connect the two.
- **Stop echoing a value that resolved to nothing.** `lookupUserByPath()` + `isInAdministeredScope()` already answer the question that matters; the only awkward part is that the failure message repeats the value back. Dropping it needs one new message key, and costs the operator the echo of what they asked for — which matters in the bulk flow.

My preference is the second, with the last path segment echoed instead of the whole value if the bulk flow needs the diagnostic. Either way the encoding on `displayName` at :255 stays — that one is free text regardless.

Not blocking this PR: the current behaviour is correct, this is about where the check belongs.

eallais:
@jayblanc your approval got dismissed by a push, sorry about that — the repo drops reviews on new commits.

Exactly one commit landed after you approved: `d6a2095`, a line wrap `prettier` asked for once the test markers got renamed. Same assertion, same file, four lines moved:

```
-  expect((win as unknown as Record<string, unknown>).__listMarkupExecuted, "…").to.be
-      .undefined
+  expect(
+      (win as unknown as Record<string, unknown>).__listMarkupExecuted,
+      "…",
+  ).to.be.undefined
```

Nothing else changed, and CI is green (13/13). Could you re-approve when you have a moment?

baptistegrimaud (approved):

PR #240, opened 2026-07-29
fix: render site settings components only for callers administering the resource

AKarmanov:
## Summary

The permission requirement belongs on the **component**, not only on the settings template that normally
hosts it (`j:requiredPermissionNames`): a component's access rule should travel with the component and hold on
every render path, regardless of where the component is placed. This module's settings screens are ordinary, instantiable content types, so the requirement is made a property of each component.

## Change

A render filter (`SettingsComponentPermissionFilter`, priority 21.5) applied on this module's six `jnt:siteSettings*` node types, registered via
**OSGi Declarative Services**. When the caller holds none of the accepted permissions it returns an empty
fragment — core's own semantics in `TemplatePermissionCheckFilter`. Because `WebflowAction` re-enters the
render chain for each webflow POST, every transition is covered too, not just the initial GET.

**The check is evaluated on the render's MAIN RESOURCE, not on the component node.** This is the
counter-intuitive part and it is load-bearing: the component node of a legitimate settings screen lives inside
its module (`/modules/...`), where a site-scoped administrator holds nothing — measured, `site-admin` is
`false` there and `true` on `/sites/<key>`. Checking the component node, which is the obvious implementation,
would refuse real site administrators. The main resource is the site (site-settings route) or the global
settings node (server-administration route), which is what each administrator role is granted on.

Either `site-admin` or `admin` is accepted: this screen is reached from both administration routes, so a
site-scoped administrator is a legitimate caller. Both are core permissions (`root-permissions.xml`) granted by
the `site-administrator` / `server-administrator` roles. The finer per-screen permissions are contributed by a
module's own `permissions.xml` and resolve to `false` where they are not registered on an instance, which would
fail closed for administrators too; the finer requirement still applies on the administration route via the
template, so this is an additional condition and never a replacement.

## Verification

Measured on 8.2.4.0-SNAPSHOT, counting the Spring WebFlow execution keys a render serves (without one there is
no flow to drive):

| caller | flow served |
|---|---|
| `root` | yes |
| server administrator | yes |
| administrator of the hosting site | **yes** — a legitimate caller for this screen |
| ordinary editor | no |

Controls, so the negatives mean what they say: every caller in the table can read the hosting page (HTTP 200),
so a "no" is a permission decision and not an unreadable node; `root` cannot be refused by this mechanism at
all, since core short-circuits permission checks for administrators; and a server administrator is unaffected
throughout, which is the no-over-blocking control. A self-contained Cypress spec asserts the same invariant in CI, with each negative paired to a positive control that goes through the identical request shape.

## Notes

- Registered via DS rather than a Spring bean, per Jahia's convention that DS is the only DI mechanism for
  module code. Both routes register into `JahiaTemplateManagerService.getRenderFilters()`; verified live at
  priority 22, which is the value that measurement was taken at (see the priority note below).
- `equals`/`hashCode` are deliberately not overridden: `AbstractFilter` defines equality as
  (concrete class, priority), which is the key `RenderService.addFilter` uses to replace an already-registered
  filter. Widening it would break that re-registration.
- Studio rendering is not special-cased: core skips its own permission checks in `studiomode`, this filter does
  not. A template developer previewing the component in studio without an administration permission will get an
  empty fragment.

- Placement is deliberately left unconstrained in the nodetype definitions. The condition holds wherever the
  component renders — which is exactly what the off-container placement in the spec exercises — so a
  definition-level restriction would be belt-and-braces rather than the load-bearing control. Recorded here so
  it does not read as an oversight.
- Priority is 21.5, not 22. An exact 22 tied with core's `templateNodeFilter` (22.0); `AbstractFilter` breaks
  such a tie on the class name, so the resulting order held by an accident of package naming. 21.5 states the
  intended slot — immediately after core's permission check at 21, clear of the 22.x template band — directly.
  No behavioural change: 22 already ordered ahead of `templateNodeFilter`.

### DIFF
--- a/src/main/java/org/jahia/modules/sitesettings/render/SettingsComponentPermissionFilter.java
+++ b/src/main/java/org/jahia/modules/sitesettings/render/SettingsComponentPermissionFilter.java
@@ -13,6 +13,23 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+/*
+ * Wiring note — this is deliberately OSGi Declarative Services, not a Spring bean.
+ *
+ * Jahia's engineering conventions (the shared `cortex` harness, skill
+ * `jahia-java-osgi-declarative-services`) state the rule as: DS is the ONLY dependency-injection
+ * mechanism allowed in a Jahia module — Blueprint is deprecated and Spring is forbidden, with the sole
+ * tolerated exception being a guarded `SpringContextSingleton.getBean(...)` read-through to a core bean.
+ * An earlier revision of this filter was registered as a Spring bean in META-INF/spring; it was moved
+ * here to follow that rule, and the render filter it registers behaves identically either way (both end
+ * up in JahiaTemplateManagerService.getRenderFilters()).
+ *
+ * The same harness documents the trap to watch for if this is ever ported to another module: without the
+ * bnd instruction `<_dsannotations>*</_dsannotations>`, an @Component class compiles and ships but emits
+ * no OSGI-INF descriptor and no Service-Component header, so the component silently never registers and
+ * the gate below simply does not run. Parents `jahia-modules` >= 8.1.7.0 switch it on already; older ones
+ * do not. Verify a descriptor is actually in the built jar rather than assuming.
+ */
 package org.jahia.modules.sitesettings.render;
 
 import org.apache.commons.lang.StringUtils;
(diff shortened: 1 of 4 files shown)


AKarmanov:
Back to draft, I need to check something.

baptistegrimaud:
@AKarmanov  , if feasible, can you backport this change onto https://github.com/Jahia/siteSettings/tree/8_7_x so it can be included in the [8.7.1 release](https://github.com/Jahia/siteSettings/milestone/4)? Thanks!

AKarmanov:
> @AKarmanov , if feasible, can you backport this change onto https://github.com/Jahia/siteSettings/tree/8_7_x so it can be included in the [8.7.1 release](https://github.com/Jahia/siteSettings/milestone/4)? Thanks!

I created a backport PR for 8.7.x: https://github.com/Jahia/siteSettings/pull/251

tests/cypress/e2e/settingsComponentRenderScope.test.cy.ts
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    const formContaining = (body: string, needle: string): string | null => {
+        const forms = body.match(/<form[^>]*>[\s\S]*?<\/form>/g) || []
+        return forms.find((f) => f.includes(needle)) ?? null
+    }
+
+    const render = (url: string) =>
+        cy
+            .request({ url, failOnStatusCode: false, qs: { ec: generateRandomID() } })
```
baptistegrimaud, line 100: `render()` forces `?ec=` (core's cache-bypass flag), so the spec only ever tests the uncached path. The cached path is the gap: the gate keys on the **main resource**, but fragments are cache-keyed on the **component node** (readable by a low-priv caller too), so an admin-populated entry could be served at the cache filter (~priority 16) before reaching this gate at 22. Core's own permission filter at 21 shares this placement, so it's probably fine via per-user keying — but CI can't show it while `ec` is forced. Worth confirming these fragments are per-user/non-cacheable, and adding one low-priv render *without* `ec` after an admin render.

tests/cypress/e2e/settingsComponentRenderScope.test.cy.ts
```diff
@@ -0,0 +1,188 @@
...
+            expect(res.body?.data?.currentUser?.name, 'the driving session must be the low-privilege account').to.eq(
+                lowPriv,
+            )
+        })
+        cy.request({ url: pageUrl, qs: { ec: generateRandomID() } }).then((res) => {
+            expect(res.status, 'the low-privilege account must be able to read the hosting page').to.eq(200)
+        })
+    })
+
+    it('serves the placed component and creates a user for the site administrator (positive control)', () => {
```
baptistegrimaud, line 168: The filter accepts `site-admin` **or** `admin`, but only `site-admin` has a positive control. The `admin` / server-administration branch (same screens under `jnt:globalSettings`) is asserted only from the manual measurement in the description — if `admin` were ever the wrong permission, this spec would stay green while server admins got locked out. Consider a second positive control granting `server-administrator` on that route, mirroring the site-admin case.

src/main/java/org/jahia/modules/sitesettings/render/SettingsComponentPermissionFilter.java
```diff
@@ -0,0 +1,137 @@
...
+ * <p>
+ * Registered via OSGi Declarative Services — no Spring context involvement.
+ */
+@Component(service = RenderFilter.class, immediate = true)
+public class SettingsComponentPermissionFilter extends AbstractFilter {
+
+    private static final Logger logger = LoggerFactory.getLogger(SettingsComponentPermissionFilter.class);
+
+    /** Node types gated by this filter. */
+    private static final String APPLY_ON_NODE_TYPES =
```
baptistegrimaud, line 87: The SEC-296 remediation asked for two things — assert the permission *and* constrain these components' placement in the nodetype definitions. This PR does the gate but not the constraint. The gate arguably subsumes it (it fires wherever the component renders, as the spec's own off-container placement shows), so this reads as a deliberate call rather than a miss — but worth recording in the PR that placement is left open on purpose, so the next reader doesn't think it was forgotten. Intended trade-off?

src/main/java/org/jahia/modules/sitesettings/render/SettingsComponentPermissionFilter.java
```diff
@@ -0,0 +1,137 @@
...
+            Collections.unmodifiableList(Arrays.asList("site-admin", "admin"));
+
+    private static final String REQUIRED_PERMISSIONS_LABEL = StringUtils.join(REQUIRED_PERMISSIONS, ", ");
+
+    @Activate
+    public void activate() {
+        // Priority 22: immediately after core's own permission check (TemplatePermissionCheckFilter, 21) and
+        // before the fragment is produced or cached — a refusal must not populate a cache entry that a
+        // differently-privileged caller could later be served.
+        setPriority(22);
```
baptistegrimaud: `AbstractFilter` priority is a float, so `22` == `22.0` — the same value as core's `templateNodeFilter` (https://github.com/Jahia/jahia-private/blob/5e201521576ec5814b58321845915c3a984892d8/core/src/main/resources/org/jahia/defaults/config/spring/applicationcontext-renderer.xml#L260). No clash today (different `applyOn`), but the tie-order is undefined. `21.5` sits unambiguously between core's permission check (`21`) and the `22.x` template/wrapper band, and makes the "immediately after 21" intent in your comment literally true.

tests/cypress/e2e/settingsComponentRenderScope.test.cy.ts
AKarmanov: Both cache filters are `applyOnModes=live`, so the administration routes never reach them; and in live the `acls` key part is a groups signature that includes the caller's own principal whenever they hold an ACE, so an entry populated by an administrator can't be served to a caller without the grant. Dropping `ec` alone wouldn't exercise that anyway (the spec renders the `default` workspace), so a real cached-path test needs publish + render-live — I'd add it separately rather than here.

tests/cypress/e2e/settingsComponentRenderScope.test.cy.ts
AKarmanov: Fair — adding a third caller granted `server-administrator` at `/`, driving the same placed component, so the `admin` branch gets a positive control through the identical request shape.

src/main/java/org/jahia/modules/sitesettings/render/SettingsComponentPermissionFilter.java
AKarmanov: Deliberate: the condition holds wherever the component renders, which is what the off-container placement in the spec demonstrates, so the definition-level constraint would be belt-and-braces. Recording it in the description so it doesn't read as an oversight.

src/main/java/org/jahia/modules/sitesettings/render/SettingsComponentPermissionFilter.java
AKarmanov: Tie order is actually deterministic — `AbstractFilter.compareTo` falls back to `getClass().getName()`, and `org.jahia.modules.…` sorts before `org.jahia.services.…`, so this already runs first. Taking `21.5` regardless since it costs nothing and makes the comment literally true; I'll apply it across the sibling modules too so they don't diverge.

baptistegrimaud (approved):
LGTM!

baptistegrimaud:
@AKarmanov , looks like the tests failed in the [PR build](https://github.com/Jahia/siteSettings/actions/runs/31003704676/job/92299029946?pr=240) 😬

PR #241, opened 2026-08-03
test: cover group member display-name rendering in Manage Groups

baptistegrimaud:
## Summary

Tests for https://github.com/Jahia/siteSettings/pull/231: Add end-to-end coverage asserting that a group member's display name renders as literal, escaped
text in Server Settings → Manage Groups, matching the existing coverage for Manage Users and page
models.

## Change

- New Cypress spec `manageGroupsDisplayEscaping.test.cy.ts`: creates a site-scoped user and group,
  adds the user as a member, and asserts the members list renders the member's display name as
  escaped text — structural sibling of `manageUsersDisplayEscaping.test.cy.ts`.
- New Groovy fixture `addSiteMemberToSiteGroup.groovy`: resolves both the group and the member
  within a given site. The existing `addMemberToGroup.groovy` fixture resolves its member with no
  site key, so it only applies to server-global members.
- The spec asserts against the `settingsBootstrap3GoogleMaterialStyle` view (visible in the webflow
  execution URL), pinning coverage to the view variant this admin route renders.

## Verification

Ran the new spec against the current module build: the members list renders the member's display
name as literal text, with no live DOM element created and no script executing.

### DIFF
--- a/tests/cypress/e2e/manageGroupsDisplayEscaping.test.cy.ts
+++ b/tests/cypress/e2e/manageGroupsDisplayEscaping.test.cy.ts
@@ -1,0 +1,89 @@
+// Regression test: the Manage Groups edit-group members list renders a member's display name as
+// inert TEXT (never live DOM/HTML), including when that member's jcr:title (self-settable by any
+// authenticated user) carries markup. Self-contained — creates its own site, group, and a member
+// user in before(), tears the site down in after(). Structural sibling of
+// manageUsersDisplayEscaping.test.cy.ts.
+//
+// This screen ships two JSP view variants for the same flow. The `.manageGroups.html` admin route
+// resolves to the `settingsBootstrap3GoogleMaterialStyle` variant in practice (confirmed empirically:
+// the webflow execution URL carries that view name, e.g.
+// `...webflowexecution<id>__settingsBootstrap3GoogleMaterialStyle=...`) — this test asserts on that
+// URL marker to pin coverage to the variant this admin route actually renders.
+import { createSite, deleteSite } from '@jahia/cypress'
+
+describe('Manage Groups - member display name rendering', () => {
+    const SITE = 'manageGroupsEscapingSite'
+    const GROUP = 'groupwithmarkupmember'
+    const USER = 'groupmarkupmember'
+    const MARKUP = '<img src=x onerror=window.__groupMembersMarkupExecuted=1>'
+
+    before(() => {
+        createSite(SITE, {
+            languages: 'en',
+            templateSet: 'templates-system',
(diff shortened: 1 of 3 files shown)


tests/cypress/e2e/manageGroupsDisplayEscaping.test.cy.ts
```diff
@@ -0,0 +1,89 @@
...
+        // open the group's members list — this is the real admin flow (submitGroupForm('editGroup', ...)),
+        // not a hand-rolled webflow POST.
+        cy.get(`a:contains(${GROUP})`, { timeout: 10000 }).first().click()
+
+        // confirm which view variant rendered — this test's coverage is specific to
+        // settingsBootstrap3GoogleMaterialStyle, not incidental to it.
+        cy.location('href').should('include', 'settingsBootstrap3GoogleMaterialStyle')
+
+        // the display name must appear as escaped literal text in a table cell
+        cy.contains('td', MARKUP, { timeout: 10000 }).should('be.visible')
```
delallioui, line 90: I think this positive assertion may be matching more cells than intended. The next column, `editGroup.jsp:226`, is `${user:fullName(member)}`, and `PrincipalViewHelper.getFullName` falls back to `escapeXml(getDisplayName(p))` when `j:firstName`/`j:lastName` are both empty — which is our case, since `createSiteUserWithTitle.groovy` only sets `jcr:title`. So that cell renders the *same* literal `MARKUP` text, and `cy.contains('td', …)` is happy with either one.

Since this is the only non-vacuous assertion here (the other two pass whenever nothing rendered at all), I'm wondering whether that makes the liveness guard a bit too loose — e.g. a later refactor that drops the username column but keeps the name column would leave this spec green while the escaping sink goes uncovered.

Would passing a `j:firstName`/`j:lastName` on the fixture user be the cheapest way to disambiguate? Then the Name cell holds `First Last` and this cell is the only one that can contain `MARKUP`. I checked `getDisplayableName()` reads `jcr:title` only, so it wouldn't disturb what you're asserting. Alternatively, anchoring to the row via its checkbox (`tr:has(input[name="selectedMembers"][value*="${USER}"])`) and then the specific cell would do it — though that's more positional. Does the ambiguity look real to you, or am I misreading which cell wins?

baptistegrimaud: yep good point, added [`55d25fb` (this PR)](https://github.com/Jahia/siteSettings/pull/241/changes/55d25fb83ae445eeb77928c942556330168b20b3)

delallioui (approved):

PR #242, opened 2026-08-03
test: add tests on site settings

sdusolle:
### Description
Add tests on site settings.
It includes test cases from Selenium test manageSiteUsers.

#### Source code

#### Tests

> [!TIP]


danielaebenberger (commented):
Thanks. Looks good.
Small improvement suggestions: 
- The new createSiteUser.groovy fixture reinvents what @jahia/cypress's createUser(name, password, props, siteKey) already does (and this same PR already imports that package for createSite/deleteSite). Worth reusing instead — Jahia/jahia-ee has independently made the same mistake, so it's a pattern worth breaking rather than repeating.
- Fixture usernames in the new createUser.test.cy.ts are hardcoded literals, not randomized — inconsistent with the sibling manageGroups.test.cy.ts added in the same PR (which does randomize), and a real flakiness risk given the new fixture silently reuses an existing user if one's left over from a failed run.

sdusolle:
Thanks for the review!

**Point 1** —  removed the duplicate createSiteUser.groovy which was a near-verbatim copy of createSiteUserWithTitle.groovy, already on main and used by the three escaping specs (identical apart from one setProperty("jcr:title", …)). So it's deleted, and the spec reuses the existing fixture with TITLE_VALUE: user. Display name stays the username, and the page objects match by path anyway (onclick*="/${username}'"), so no selector changes.

**Point 2** — randomized usernames: fixed. All eight fixture usernames now carry a per-run uniq, matching the sibling manageGroups.test.cy.ts:

```
const uniq = generateRandomID().replace(/[^a-z0-9]/gi, '')
const EXISTING_USER = 'existingTestUser' + uniq
```

danielaebenberger (approved):

PR #243, opened 2026-08-03
fix: consistently escape the page-model path in the Page Models admin views

baptistegrimaud:
## Summary

On the Page Models administration screen, a page-model's path is rendered in several
HTML-attribute positions on each row, in addition to the visible link text. This change escapes
the path consistently in every one of those positions, so a path containing reserved characters
(such as a quote or an angle bracket) is always emitted as a well-formed attribute value.

## Change

- `siteSettingsManagePageModels.jsp` and its `settingsBootstrap3GoogleMaterialStyle` variant: wrap
  `pageModel.path` in `fn:escapeXml` in the three attribute-context occurrences on each row — the
  module-hook `<div>`'s `path="…"` attribute and the two row-link `href="<c:url …/>"` values. This
  aligns those occurrences with the visible link text (already rendered through `<c:out>`) and with
  the display name in the same view (already rendered through `fn:escapeXml`). The escaped value is
  computed once per row in a `<c:set var="pageModelPathAttr">` and referenced from all three
  positions, so a new attribute added to the row inherits the same handling.
- The `<c:url>` on the two row links is kept: it prepends the context path and applies
  `response.encodeURL()` (session rewriting) to the value — it does **not** percent-encode the value
  itself (that applies to `<c:param>` name/value pairs only). `fn:escapeXml` is applied in addition,
  to encode the attribute value. The two are independent concerns — `<c:url>` composes the link,
  `fn:escapeXml` makes it a well-formed attribute value. (URL-encoding reserved characters within the
  path is a separate concern this change does not address.)

## Verification

- Added a self-contained Cypress spec (`managePageModelsPathAttributeEscaping.test.cy.ts`, with a
  small Groovy fixture) that plants a template-model page whose node name contains reserved markup
  characters and asserts the Page Models screen renders the path as an inert attribute value — no
  injected element and no handler execution.
- Ran the module's e2e suite against a local Jahia 8.2 instance: both the new spec and the existing
  page-models display-escaping spec pass, including after the escaping/test-setup consolidation
  (shared row-wait helper, single Groovy fixture).


src/main/resources/jnt_siteSettingsManagePageModels/html/siteSettingsManagePageModels.jsp
```diff
@@ -51,7 +51,7 @@
         <tr class="${status.index % 2 == 0 ? 'evenLine' : 'oddLine'}">
             <td>
                 <div class="jahia-template-gxt" jahiatype="module" id="module${pageModel.identifier}" type="existingNode"
-                     scriptInfo="" path="${pageModel.path}" template="hidden.system" dragdrop="false">
+                     scriptInfo="" path="${fn:escapeXml(pageModel.path)}" template="hidden.system" dragdrop="false">
```
eallais: The escaping itself is right, but it's now spelled out at each of the three attribute occurrences in this view and again in the `settingsBootstrap3GoogleMaterialStyle` variant — six copies of one decision. The `<c:forEach>` body gives you a natural place to state it once:

```jsp
<c:set var="pageModelPathAttr" value="${fn:escapeXml(pageModel.path)}"/>
```

then `path="${pageModelPathAttr}"` here, and `<c:url value='${url.base}${pageModelPathAttr}.html'/>` on the two row links.

One rule per row instead of three independent ones: the next attribute added to this row inherits it, rather than depending on a reviewer noticing a bare `${pageModel.path}`.

**Not blocking — the current form is correct. This is about making the next occurrence safe by default.**

**Source:** the sibling variant carries the same three occurrences one line off — `siteSettingsManagePageModels.settingsBootstrap3GoogleMaterialStyle.jsp:55,64,67`.

tests/cypress/fixtures/groovy/createPageModelWithEncodedName.groovy
```diff
@@ -0,0 +1,45 @@
...
+// then rename()-ing to the decoded name.
+//
+// Runs in the EDIT workspace + EN locale: the screen renders in edit/en and its list query joins the EN
+// translation, so the node must carry its mandatory i18n props (jcr:title, j:pageTemplateTitle) in EN or
+// it is filtered out of the list. Idempotent. Tokens replaced by cy.executeGroovy.
+JCRTemplate.getInstance().doExecuteWithSystemSessionAsUser(null, Constants.EDIT_WORKSPACE, Locale.ENGLISH, new JCRCallback() {
+    @Override
+    Object doInJCR(JCRSessionWrapper session) throws RepositoryException {
+        JCRNodeWrapper site = session.getNode("/sites/SITE_KEY")
+        String markupName = new String(Base64.getDecoder().decode("MARKUP_NAME_B64"), "UTF-8")
```
eallais, line 22: This fixture is `createPageModelWithMarkupName.groovy` plus a base64 hop, and that hop is a strict superset — it carries any node name, quotes included. Everything else is identical down to the comments; the older one even still logs `createPageModelWithNamePayload:`.

- Worth deleting the older fixture and pointing `managePageModelsDisplayEscaping.test.cy.ts` at this one with `MARKUP_NAME_B64: btoa(MARKUP)`.
- One fixture then owns "plant a template-model page under a name I choose", so the plant recipe has a single place to drift — the EDIT/EN session, the mandatory `j:pageTemplateTitle`, the `pageModelProbe` cleanup.

**Source:** `tests/cypress/fixtures/groovy/createPageModelWithMarkupName.groovy:22` (`String markupName = "MARKUP_NAME"`) and `:39` (the stale `createPageModelWithNamePayload` log line).

tests/cypress/e2e/managePageModelsPathAttributeEscaping.test.cy.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    after(() => {
+        deleteSite(SITE)
+    })
+
+    // Open the Page Models admin screen and reload until the planted page-model surfaces in the list.
+    // The row is populated by a JCR-SQL2 query served from Oak's async index, which lags a fresh plant
+    // by a few seconds; without this the list can render empty and the assertions would be vacuous.
+    const MAX_ATTEMPTS = 10
+    const openUntilRowPresent = (attempt = 0): Cypress.Chainable => {
```
eallais: This helper — and the `before`/`after` around it — is character-for-character the one in `managePageModelsDisplayEscaping.test.cy.ts`, the window-flag name aside. It's the slowest and most fragile part of both specs, now maintained twice.

- Consider lifting it into `tests/cypress/support`, taking the flag name as an argument, so the index-lag pacing (`MAX_ATTEMPTS`, the 3s step) is tuned in one place.
- Going one step further: this spec's payload already leads with `">`, so it exercises the text-node column too. The two specs could plant one page on one site and assert both properties — that drops one `createSite` and one retry loop, which is where the wall-clock goes.

Both are cleanups rather than objections — fine by me if you'd rather keep the specs independent.

**Source:** `tests/cypress/e2e/managePageModelsDisplayEscaping.test.cy.ts:39-62` — same `MAX_ATTEMPTS`, same `cy.get('#pageModelsTable tbody', { timeout: 10000 })`, same 3s `cy.wait` with the same eslint-disable.

src/main/resources/jnt_siteSettingsManagePageModels/html/siteSettingsManagePageModels.jsp
```diff
@@ -60,9 +60,9 @@
                 </div>
             </td>
             <td>
-                <a href="<c:url value='${url.base}${pageModel.path}.html'/>"><c:out value='${pageModel.properties["j:pageTemplateTitle"].string}'/></a>
+                <a href="<c:url value='${url.base}${fn:escapeXml(pageModel.path)}.html'/>"><c:out value='${pageModel.properties["j:pageTemplateTitle"].string}'/></a>
             </td>
-            <td>  <a href="<c:url value='${url.base}${pageModel.path}.html'/>"><c:out value='${pageModel.path}'/></a></td>
+            <td>  <a href="<c:url value='${url.base}${fn:escapeXml(pageModel.path)}.html'/>"><c:out value='${pageModel.path}'/></a></td>
```
eallais: One small correction to the description's split of duties: `<c:url>` prepends the context path and runs `response.encodeURL()`, but it does not percent-encode `value` — `URLEncoder.encode` is only applied to `<c:param>` name/value pairs. So the attribute value is well-formed markup, while the URL inside it is not a valid URL: a node name carrying a space, a `#`, or a `?` reaches the browser as-is and the link resolves elsewhere, or nowhere.

- Independent of what this PR is doing, so no objection to leaving it.
- Deliberately out of scope? If a working link matters for those names, the value needs per-segment URL-encoding (the `/` separators have to survive) *before* the XML escape — a different change from this one.

**Source:** `apache/tomcat-taglibs-standard` — `UrlSupport.resolveUrl` only does `result = response.encodeURL(result)`; the percent-encoding lives in `ParamSupport.java:99` (`URLEncoder.encode(name, enc), URLEncoder.encode(value, enc)`), i.e. `<c:param>` only.

tests/cypress/fixtures/groovy/createPageModelWithEncodedName.groovy
```diff
@@ -0,0 +1,45 @@
...
+import javax.jcr.RepositoryException
+import java.util.Base64
+import java.util.Locale
+
+// Plant a template-model page (jnt:page + jmix:canBeUseAsTemplateModel) under the site whose NODE NAME
+// is supplied base64-encoded. The encoding is what lets the name carry characters — a double quote in
+// particular — that would otherwise terminate this script's own string literal when the token is
+// substituted in; the name is decoded here so it reaches the repository verbatim. That node name flows
+// into the "page path" column of the Page Models administration screen. A node name set through addChild
+// is sanitised, whereas a JCR rename() is not; the fixture reproduces that by creating with a plain name
```
eallais: This header's closing sentence is written as a contrast between two platform behaviors — one API does A, another doesn't. On a public repo a fixture comment carries less if it states only what the fixture itself does:

```groovy
// Creates the node under a plain name, then renames it to the decoded name.
```

- Same for the "why base64" sentence just above it: `// the name is passed base64-encoded so it can carry any character a JCR name allows` says everything the next reader needs.
- The spec header at `managePageModelsPathAttributeEscaping.test.cy.ts:1-6` reads the same way — it explains a rendering failure mode rather than the property being asserted. `// Asserts the page-model path renders as an inert attribute value on the Page Models screen.` covers it.
- No test cost either way: nothing here changes what runs, only what the comments narrate. Worth doing before the branch is cherry-picked, since the same files travel to #244.

src/main/resources/jnt_siteSettingsManagePageModels/html/siteSettingsManagePageModels.jsp
baptistegrimaud: agreed, it makes sense, done in [`781d2d8` (this PR)](https://github.com/Jahia/siteSettings/pull/243/changes/781d2d8bb43e66f9e3fc9e46a2b742b9d321e97e)

tests/cypress/fixtures/groovy/createPageModelWithEncodedName.groovy
baptistegrimaud: Done in [`781d2d8` (this PR)](https://github.com/Jahia/siteSettings/pull/243/changes/781d2d8bb43e66f9e3fc9e46a2b742b9d321e97e). Deleted createPageModelWithMarkupName.groovy and pointed managePageModelsDisplayEscaping.test.cy.ts at this fixture with MARKUP_NAME_B64: btoa(MARKUP). One fixture now owns the plant recipe — the EDIT/EN session, the mandatory j:pageTemplateTitle, the pageModelProbe cleanup — and the stale createPageModelWithNamePayload: log line went with it. Both specs pass against a local 8.2 instance after the change.

tests/cypress/e2e/managePageModelsPathAttributeEscaping.test.cy.ts
baptistegrimaud: Took the first half: lifted the reload-until-row-present helper (with its index-lag pacing — `MAX_ATTEMPTS`, the 3s step) into `cypress/support/pageModels.ts` as `openPageModelsUntilRow(siteKey, handlerFlag)`, and both specs import it, so that fragile pacing now lives in one place. I kept the two specs independent rather than folding them onto one site: they cover two different rendering positions (the text cell vs. the row's attributes), and I'd rather each remain a standalone regression than share one setup. That does keep the extra `createSite` + retry loop — happy to revisit if the wall-clock becomes a concern.

src/main/resources/jnt_siteSettingsManagePageModels/html/siteSettingsManagePageModels.jsp
baptistegrimaud: Correct, and thanks for the precision: `<c:url>` prepends the context path and runs `response.encodeURL()` but does not percent-encode `value` — that is `<c:param>` only. So a node name carrying a space, `#` or `?` still yields a link that resolves elsewhere. That is a pre-existing behaviour this change neither introduces nor worsens (attribute-value handling is orthogonal to URL validity), so I've deliberately left it out of scope — a working link for those names needs per-segment URL-encoding *before* the escape, which is a distinct change. I'll also tighten the PR description's "URL transformation" wording, which overstated what `<c:url>` does.

tests/cypress/fixtures/groovy/createPageModelWithEncodedName.groovy
baptistegrimaud: Agreed — trimmed this header in [`781d2d8` (this PR)](https://github.com/Jahia/siteSettings/pull/243/changes/781d2d8bb43e66f9e3fc9e46a2b742b9d321e97e) and both spec headers to state only what they do (plant a page under a chosen node name; assert how the path renders), dropping the platform-behaviour contrast and the rendering-mechanics narration. The base64 rationale is reduced to "so the name can carry any character a JCR name allows".

baptistegrimaud: Once this PR is merged, I'll apply the same change onto https://github.com/Jahia/siteSettings/pull/244

eallais (approved):

PR #244, opened 2026-08-03
fix: consistently escape the page-model path in the Page Models admin views

baptistegrimaud:
Backport #243 onto https://github.com/Jahia/siteSettings/tree/8_7_x

Bottom of a stack with #245: this PR escapes the page-model path in its **attribute** positions (the `<div>` `path=` and the two `<c:url>` `href=`); the sibling #245, stacked on top, escapes the adjacent **text** values (display name, template title, path link). Both target `8_7_x` under milestone `siteSettings 8.7.1` and are released together — no 8.7.x build is cut between them.

_Tests are not backported: this maintenance line has no `tests/` Cypress harness (none existed at 8.7.0), so the dev-line regression spec has no counterpart here._

_The Static Analysis check is red on a pre-existing `audit-ci`/Node tooling issue on this maintenance branch, unrelated to this change (the diff is JSP-only)._

### DIFF
--- a/src/main/resources/jnt_siteSettingsManagePageModels/html/siteSettingsManagePageModels.jsp
+++ b/src/main/resources/jnt_siteSettingsManagePageModels/html/siteSettingsManagePageModels.jsp
@@ -51,7 +51,7 @@
         <tr class="${status.index % 2 == 0 ? 'evenLine' : 'oddLine'}">
             <td>
                 <div class="jahia-template-gxt" jahiatype="module" id="module${pageModel.identifier}" type="existingNode"
-                     scriptInfo="" path="${pageModel.path}" template="hidden.system" dragdrop="false">
+                     scriptInfo="" path="${fn:escapeXml(pageModel.path)}" template="hidden.system" dragdrop="false">
                     <table class="table table-bordered">
                         <tr>
                             <td>${pageModel.displayableName}</td>
@@ -60,9 +60,9 @@
                 </div>
             </td>
             <td>
-                <a href="<c:url value='${url.base}${pageModel.path}.html'/>">${pageModel.properties["j:pageTemplateTitle"].string}</a>
+                <a href="<c:url value='${url.base}${fn:escapeXml(pageModel.path)}.html'/>">${pageModel.properties["j:pageTemplateTitle"].string}</a>
             </td>
-            <td>  <a href="<c:url value='${url.base}${pageModel.path}.html'/>">${pageModel.path}</a></td>
+            <td>  <a href="<c:url value='${url.base}${fn:escapeXml(pageModel.path)}.html'/>">${pageModel.path}</a></td>
 
         </tr>
         <c:set var="isEmpty" value="false"/>
--- a/src/main/resources/jnt_siteSettingsManagePageModels/html/siteSettingsManagePageModels.jsp
+++ b/src/main/resources/jnt_siteSettingsManagePageModels/html/siteSettingsManagePageModels.jsp
@@ -48,10 +48,11 @@
     <tbody>
     <c:forEach items="${moduleMap.currentList}" var="pageModel" begin="${moduleMap.begin}" end="${moduleMap.end}"
                varStatus="status">
+        <c:set var="pageModelPathAttr" value="${fn:escapeXml(pageModel.path)}"/>
         <tr class="${status.index % 2 == 0 ? 'evenLine' : 'oddLine'}">
             <td>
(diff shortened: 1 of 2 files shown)


src/main/resources/jnt_siteSettingsManagePageModels/html/siteSettingsManagePageModels.jsp
```diff
@@ -60,9 +60,9 @@
                 </div>
             </td>
             <td>
-                <a href="<c:url value='${url.base}${pageModel.path}.html'/>">${pageModel.properties["j:pageTemplateTitle"].string}</a>
+                <a href="<c:url value='${url.base}${fn:escapeXml(pageModel.path)}.html'/>">${pageModel.properties["j:pageTemplateTitle"].string}</a>
             </td>
-            <td>  <a href="<c:url value='${url.base}${pageModel.path}.html'/>">${pageModel.path}</a></td>
+            <td>  <a href="<c:url value='${url.base}${fn:escapeXml(pageModel.path)}.html'/>">${pageModel.path}</a></td>
```
eallais: The `href` is now escaped, but the link text right next to it still isn't — and on `8_7_x` that's true of the whole row, not just this cell.

- On `main` these three sibling values were already routed through `<c:out>` / `fn:escapeXml` before #243 landed, which is why #243's diff shows them as untouched context: `<c:out value='${pageModel.path}'/>` here, `<c:out value='${pageModel.properties["j:pageTemplateTitle"].string}'/>` on line 63, and `${fn:escapeXml(pageModel.displayableName)}` on line 57.
- That earlier pass was never carried to this line, so on `8_7_x` all three are still raw `${…}`. This PR escapes the attribute positions around them and leaves the values themselves as they were.
- [#245](https://github.com/Jahia/siteSettings/pull/245) is based on *this* branch and does cover them, so the pair is complete — but nothing on this PR says so, and its own title promises the path is escaped "consistently", which only becomes true once #245 is in.

**Worth saying it in the body: this is the bottom of a stack with #245 and the two go to `8_7_x` together, so no 8.7.x build is cut between them.** Same picture in the `settingsBootstrap3GoogleMaterialStyle` variant (lines 58, 64, 67).

**Source:** [`siteSettings` #243 diff](https://github.com/Jahia/siteSettings/pull/243/files) — the same three lines appear there as unchanged context already carrying `<c:out>` / `fn:escapeXml`; on this branch `grep -rn 'pageModel\.' src/main/resources/jnt_siteSettingsManagePageModels/` shows them raw.

src/main/resources/jnt_siteSettingsManagePageModels/html/siteSettingsManagePageModels.jsp
```diff
@@ -60,9 +60,9 @@
                 </div>
             </td>
             <td>
-                <a href="<c:url value='${url.base}${pageModel.path}.html'/>">${pageModel.properties["j:pageTemplateTitle"].string}</a>
+                <a href="<c:url value='${url.base}${fn:escapeXml(pageModel.path)}.html'/>">${pageModel.properties["j:pageTemplateTitle"].string}</a>
```
eallais: Escaping inside `<c:url>`'s `value` puts the markup-escaping step *before* the URL is assembled, which works here but only because `url.base` happens to be entity-free.

- The usual order is the other way round — build the URL, then escape the whole attribute value on output — so nothing in the concatenation can slip through unescaped no matter what the base contributes:

```jsp
<c:url var="pageModelUrl" value="${url.base}${pageModel.path}.html"/>
<a href="${fn:escapeXml(pageModelUrl)}">…</a>
```

- Not offering this as a one-click suggestion on purpose: it would collide with #245, which edits the same line.
- It also isn't worth diverging from `main` for. If it's worth doing, it's worth doing on #243 first so the two lines stay identical and the next backport is still a clean cherry-pick.

**Source:** self-evident from the diff; `url.base` is server-built (`org.jahia.services.render.URLGenerator`), which is why the current form is safe in practice.

eallais (changes requested):

src/main/resources/jnt_siteSettingsManagePageModels/html/siteSettingsManagePageModels.jsp
baptistegrimaud: Both PRs are [stacked together](https://docs.github.com/en/pull-requests/how-tos/stacked-pull-requests) and part of the same milestone.
The 8.7.1 release will indeed contain both

eallais (dismissed):

eallais (approved):

PR #248, opened 2026-08-03
fix: bind the user and group management screens to the realm of their container

eallais:
## Summary

The Manage Users and Manage Groups administration screens administer the principals of one realm.
Two containers carry a realm: the global settings node, whose realm is the server-wide one and whose
store is the server-global `/users` and `/groups`, and a site node, whose realm is that site and
whose store is `/sites/<siteKey>/…`. A container of any other type carries no realm, and these
screens treat that as a realm they cannot serve.

## Change

`ManageGroupsFlowHandler` and `UsersFlowHandler` each record whether `initRealm` recognized the
container, alongside the site key it resolves:

- `initRealm` recognizes `jnt:virtualsite` (a site realm, site key set) and `jnt:globalSettings`
  (the server-wide realm, no site key). Any other container leaves the realm unresolved.
- `isInAdministeredScope`, `isAllowedMember` and `isSiteKeyInScope` answer `false` while the realm is
  unresolved, so every target, member and destination check declines — covering `addGroup`,
  `addMembers`, `removeMembers`, `removeGroup`, `copyGroup`, `updateUser` and `removeUser`.
- `addUser` and `bulkAddUser`, which take the resolved site key as the store to create in, decline
  with the existing `siteSettings.user.create.unsuccessful` message.

The site key alone cannot carry this distinction, because the server-wide realm legitimately has
none: the container's own type is what identifies it. The realm is read from that type rather than
inferred from an enclosing site, so the site key a screen holds is only ever the realm it is bound
to.

**Behaviour boundary.** A deployment that holds one of these screens in any other container sees the
screen list principals and apply no change. Both containers the screens are shipped in — the site
settings page and the server settings page — are unaffected.

## Verification

`tests/cypress/e2e/Groups/manageGroupsRealm.test.cy.ts` drives the three containers through the
screen's own web flow: the server-wide realm (`/settings`) and a site realm each add a member to a
group of their own store, and a container carrying no realm adds none. The two realm cases are the
positive controls for the third — same flow, same transition, same read-back — so a driver that had
stopped working, or a read-back looking in the wrong place, would take them red too. Group members
are read at full depth, several levels below the group, where a children-only query reads a
successful write as an absent one.

Against a local 8.2.4.0-SNAPSHOT instance with this module deployed:

- the three new cases pass, and fail on the third alone without this change;
- the existing `manageGroups*` specs stay green (14/14);
- two `Users` specs fail on that instance for want of `dx-base-demo-templates` and an unrelated
  fixture script — identically with and without this change.

### DIFF
--- a/src/main/java/org/jahia/modules/sitesettings/users/UsersFlowHandler.java
+++ b/src/main/java/org/jahia/modules/sitesettings/users/UsersFlowHandler.java
@@ -56,14 +56,29 @@ public class UsersFlowHandler implements Serializable {
 
     private String siteKey;
 
+    private boolean realmResolved;
+
     private transient JahiaPasswordPolicyService pwdPolicyService;
 
     private transient JahiaUserManagerService userManagerService;
 
+    /**
+     * Resolves the principal realm this screen manages from the container it is reached through: a site
+     * node scopes it to that site's principals, the global settings node to the server-global store
+     * (hence a null {@link #siteKey}). Any other container carries no realm, leaving
+     * {@link #realmResolved} false — the state in which every scope check below answers false and the
+     * creation paths that take {@link #siteKey} as their destination decline.
+     */
     public void initRealm(RenderContext renderContext) throws RepositoryException {
         JCRNodeWrapper mainNode = renderContext.getMainResource().getNode();
-        if (mainNode != null && mainNode.isNodeType("jnt:virtualsite")) {
+        if (mainNode == null) {
+            return;
+        }
+        if (mainNode.isNodeType("jnt:virtualsite")) {
             siteKey = ((JCRSiteNode) mainNode).getSiteKey();
+            realmResolved = true;
+        } else if (mainNode.isNodeType("jnt:globalSettings")) {
+            realmResolved = true;
         }
     }
 
@@ -73,9 +88,10 @@ public class UsersFlowHandler implements Serializable {
      * site's principal tree (/sites/&lt;siteKey&gt;/...); in the server-wide realm (siteKey null) the
      * target must live in the server-global store (/users/ or /groups/) and never inside a site's tree.
      * This mirrors the store layout used by JahiaUserManagerService ("/users/" vs "/sites/&lt;siteKey&gt;/users/").
+     * With no realm resolved at all nothing is in scope.
      */
     private boolean isInAdministeredScope(String principalPath) {
-        if (principalPath == null) {
+        if (principalPath == null || !realmResolved) {
             return false;
         }
         if (siteKey == null) {
@@ -86,6 +102,10 @@ public class UsersFlowHandler implements Serializable {
 
     public boolean addUser(final UserProperties userProperties, final MessageContext context) throws RepositoryException {
         logger.info("Adding user");
+        if (!realmResolved) {
+            context.addMessage(new MessageBuilder().error().code("siteSettings.user.create.unsuccessful").build());
+            return false;
+        }
         return JCRTemplate.getInstance().doExecuteWithSystemSession(new JCRCallback<Boolean>() {
             @Override
             public Boolean doInJCR(JCRSessionWrapper session) throws RepositoryException {
@@ -120,6 +140,10 @@ public class UsersFlowHandler implements Serializable {
 
     public boolean bulkAddUser(final CsvFile csvFile, final MessageContext context) throws RepositoryException {
         logger.info("Bulk adding users");
+        if (!realmResolved) {
+            context.addMessage(new MessageBuilder().error().code("siteSettings.user.create.unsuccessful").build());
+            return false;
+        }
 
         long timer = System.currentTimeMillis();
         boolean hasErrors = JCRTemplate.getInstance().doExecuteWithSystemSession(new JCRCallback<Boolean>() {
--- a/src/main/java/org/jahia/modules/sitesettings/users/UsersFlowHandler.java
+++ b/src/main/java/org/jahia/modules/sitesettings/users/UsersFlowHandler.java
@@ -281,7 +281,17 @@ public class UsersFlowHandler implements Serializable {
         });
     }
 
+    /**
+     * Lists the principals of the administered realm. The realm is what bounds the listing: a site realm
+     * lists that site's store, the server-wide realm the server-global one. With no realm resolved there
+     * is no store to list, so the result is empty rather than the server-global default a null
+     * {@link #siteKey} would otherwise select.
+     */
     public Set<JCRUserNode> search(SearchCriteria searchCriteria) {
+        if (!realmResolved) {
+            searchCriteria.setNumberOfRemovedJahiaAdministrators(0);
+            return Collections.emptySet();
+        }
         String searchTerm = searchCriteria.getSearchString();
         if (StringUtils.isNotEmpty(searchTerm) && searchTerm.indexOf('*') == -1) {
             searchTerm += '*';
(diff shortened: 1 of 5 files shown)


src/main/java/org/jahia/modules/sitesettings/users/UsersFlowHandler.java
```diff
@@ -86,6 +102,10 @@ private boolean isInAdministeredScope(String principalPath) {
 
     public boolean addUser(final UserProperties userProperties, final MessageContext context) throws RepositoryException {
         logger.info("Adding user");
+        if (!realmResolved) {
```
baptistegrimaud, line 105: Any reasons to not add this check for `search()`, `init()` and `getProvidersList()` as well?

baptistegrimaud (commented):
just a minor question, looks good otherwise

baptistegrimaud:
@eallais , if feasible, can you backport this change onto https://github.com/Jahia/siteSettings/tree/8_7_x so it can be included in the [8.7.1 release](https://github.com/Jahia/siteSettings/milestone/4)? Thanks!

src/main/java/org/jahia/modules/sitesettings/users/UsersFlowHandler.java
eallais: Bound \`search()\` and \`getProvidersList()\` in 8a093cd: with no realm resolved both now answer empty, and the withheld-principal count resets — previously the listing fell through to the store a null realm key selects.

Left \`init()\` as it is, deliberately: it is evaluated by neither flow definition nor any JSP, so it is not on the render path. The `search` view-state's \`on-render\` is what populates the listing on every render. A guard in \`init()\` would be a branch nothing reaches, and it would suggest to the next reader that the method is live.

Verified on a running instance rather than by inspection: the new listing spec's three cases pass, and its third case fails on a build without the bound listing — so it discriminates rather than merely passing. The existing Manage Users and Manage Groups specs stay green.

baptistegrimaud (approved):

eallais:
@baptistegrimaud Feasible, but bigger than it looks on that branch — and the scope call is yours since you own the 8.7.1 release.

`8_7_x` does not carry the earlier scope work. Measured: the four commits of #226 and #227 are **not** ancestors of `8_7_x`, and `isInAdministeredScope` / `isAllowedMember` / `isSiteKeyInScope` are absent from both handlers there. #248 refines exactly those, so on its own it does not apply — cherry-picking it onto `8_7_x` conflicts in both `UsersFlowHandler` and `ManageGroupsFlowHandler`.

What I measured on a throwaway branch off `8_7_x`:

| step | result |
|---|---|
| the 4 commits of #226 + #227 | apply **cleanly**, and bring their own self-contained spec (`siteAdminPrincipalScope.test.cy.ts`) |
| #248 on top | conflicts on both handlers — resolvable by hand, since it refines what the 4 just added |
| the 2 newer regression specs | **cannot** come as-is: `8_7_x` has no `tests/cypress/utils/`, no `page-object/` and no `fixtures/groovy/`, so they import and call things that do not exist on the line |

So a backport that actually holds on `8_7_x` means #226 + #227 + #248 together, and the two newer specs rewritten self-contained the way the line’s own spec is.

Two questions before I open anything:

1. Do you want that full scope in 8.7.1, or Java only, keeping the line’s existing spec as the coverage?
2. If Java only, the newer behaviour ships on that line without its own regression — fine by you?

The branch is prepared locally either way; tell me which shape and I will open the PR against `8_7_x`.

eallais:
Correcting my previous comment on two points. The conclusion changes: the backport applies cleanly, and the trap is elsewhere than I said.

**1. I was wrong that #248 does not apply.** My commit list was incomplete. I built it with `git log -S"isInAdministeredScope"`, which only reports commits where the *count* of that name changes, so it silently skipped `651fec4` ("Also constrain the server-wide realm to the global principal store (/users, /groups)") from #226 — a commit that edits the method body without touching its name. With `651fec4` included, all seven commits cherry-pick onto `8_7_x` **with no conflict at all**:

`38b8e4c` → `651fec4` → `f3079e7` → `833a85e` (#226) → `c2780f2` (#227) → `aebe783` (#248) → `8a093cd` (#248)

#239 is not needed for this.

**2. The real trap, and it is worth knowing before anyone else touches this line.** My earlier "the four commits apply cleanly" was true and misleading. Without `651fec4`, those four apply with **no conflict** and leave `isInAdministeredScope` as:

```java
if (siteKey == null) {
    return true;
}
```

where the same four commits on `main` leave it as:

```java
if (siteKey == null) {
    return principalPath.startsWith("/users/") || principalPath.startsWith("/groups/");
}
```

So a clean cherry-pick produced a weaker check than the one `main` carries, with nothing to signal it — it compiles, and no conflict marker appears. With the full seven, the line ends up byte-identical to `main` on both guard methods. Please keep `651fec4` in any ordering of this backport.

**3. The specs: Java only, by necessity rather than by choice.** `8_7_x` has **zero files under `tests/`** — no Cypress project on the line at all. The seven commits do bring three specs with them, but they cannot run there and their imports (`utils`, page objects, Groovy fixtures) resolve to nothing. I dropped them in a separate commit rather than ship dead files. So the net backport is 2 Java files plus 3 changelog fragments.

**4. One thing you may want to know for the 8.7.1 build.** The line pins `frontend-maven-plugin` at `nodeVersion v14.16.0`, which 404s on Apple Silicon — Node has no arm64 macOS build before 16 — so `mvn compile` cannot run on an M-series Mac. I verified the Java instead by resolving the dependency classpath and compiling with `javac` directly: 7 sources, clean.

Branch prepared and ready. Say the word and I open it against `8_7_x`.

baptistegrimaud:
> @baptistegrimaud Feasible, but bigger than it looks on that branch — and the scope call is yours since you own the 8.7.1 release.
> 
> `8_7_x` does not carry the earlier scope work. Measured: the four commits of #226 and #227 are **not** ancestors of `8_7_x`, and `isInAdministeredScope` / `isAllowedMember` / `isSiteKeyInScope` are absent from both handlers there. #248 refines exactly those, so on its own it does not apply — cherry-picking it onto `8_7_x` conflicts in both `UsersFlowHandler` and `ManageGroupsFlowHandler`.
> 
> What I measured on a throwaway branch off `8_7_x`:
> 
> step	result
> the 4 commits of #226 + #227	apply **cleanly**, and bring their own self-contained spec (`siteAdminPrincipalScope.test.cy.ts`)
> #248 on top	conflicts on both handlers — resolvable by hand, since it refines what the 4 just added
> the 2 newer regression specs	**cannot** come as-is: `8_7_x` has no `tests/cypress/utils/`, no `page-object/` and no `fixtures/groovy/`, so they import and call things that do not exist on the line
> So a backport that actually holds on `8_7_x` means #226 + #227 + #248 together, and the two newer specs rewritten self-contained the way the line’s own spec is.
> 
> Two questions before I open anything:
> 
> 1. Do you want that full scope in 8.7.1, or Java only, keeping the line’s existing spec as the coverage?
> 2. If Java only, the newer behaviour ships on that line without its own regression — fine by you?
> 
> The branch is prepared locally either way; tell me which shape and I will open the PR against `8_7_x`.

Thanks for looking into that.
It's fine to not backport the Cypress tests as the setup in place. That's what I did for https://github.com/Jahia/siteSettings/pull/244 and https://github.com/Jahia/siteSettings/pull/245.

If the resulting branch with the different cherry-picks is relatively close to the `main` branch, let's open the PR and include it in 8.7.1. If they diverge too much, we can skip the backport as long as we document that decision.

PR #249, opened 2026-08-04
fix: scope the user and group management screens to the realm they administer

eallais:
Backport onto `8_7_x` for the 8.7.1 milestone, requested in #248.

Java only — no Cypress specs, matching what #244 and #245 did for this line. `8_7_x` carries no `tests/` tree at all, so the specs that travel with these commits cannot run there and their imports resolve to nothing. They are removed in a dedicated commit.

## What is in it

`8_7_x` did not carry the earlier scope work: `isInAdministeredScope`, `isAllowedMember` and `isSiteKeyInScope` were absent from both handlers. So this brings the whole family, seven commits, cherry-picked in order and all applying without conflict:

| commit | from |
|---|---|
| `38b8e4c` enforce site scope in the two administration flows | #226 |
| `651fec4` constrain the server-wide realm to the global principal store | #226 |
| `f3079e7` keep the target-group check, allow global principals as members | #226 |
| `833a85e` gate cross-site principals as group members | #226 |
| `c2780f2` scope create / copy / remove-members in Manage Groups | #227 |
| `aebe783` bind both screens to the realm of their container | #248 |
| `8a093cd` list principals only for a realm the container resolves | #248 |

**`651fec4` is load-bearing and easy to miss.** Without it the other six still apply with no conflict, and `isInAdministeredScope` ends as `if (siteKey == null) { return true; }` — a weaker check than the one `main` carries, with nothing to signal it. Keep it in any reordering.

Two commits from review follow the cherry-picks: `6826bb1` applies to the unresolved-path argument of `removeUser` the encoder `main` applies to it, and `a39f9e1` drops the changelog fragments (see below).

## How close is this to `main`

Both guard methods are **byte-identical** to `main`. What still differs in the two files is pre-existing divergence of the line, unrelated to this change:

- `UsersFlowHandler`, 22 lines — `CsvFile` exposes its separator as a `char` here and as a `String` on `main`; the `escapeXml` calls on message arguments come from #239, which is not part of what #248 asked for and is not backported here — except the one on `removeUser`'s unresolved-path argument, which the scope branch introduces on this line and which therefore follows `main` (`6826bb1`). The remaining ones cover values `8_7_x` already reported raw before this backport, so they belong to a backport of #239 rather than to this one. One later `j:invalidateSessionTime` line is absent.
- `ManageGroupsFlowHandler`, 40 lines — older membership idioms (`members.contains(...)` where `main` uses `groupNode.isMember(...)` and `user.isMemberOfGroup(...)`), plus whitespace.

## Release notes

The three `.chachalog/` fragments that travelled with the cherry-picks are dropped in `a39f9e1`: `8_7_x` has no `.chachalog/` directory and none of the `chachalog-*` workflows, so nothing on this line consumes them, and #244 carries none either. Nothing therefore generates an 8.7.1 entry for this change — whoever writes those release notes wants this line:

> The Manage Users and Manage Groups administration screens are now bound to the principal realm of the container they are reached through: the server-wide realm on the global settings node, a single site's realm on a site node. Held by any other container, these screens apply no change; Manage Users additionally lists no principal.

The wording is deliberately narrower than the fragment's on `main`: the read side of Manage Groups was never gated, so that screen still enumerates the server-global groups and providers in an unresolved realm. That gap exists identically on `main` and is a follow-up there, not a regression of this backport.

## Verification

`javac` over the module's 7 sources against the resolved dependency classpath: clean, both handler classes produced.

`mvn compile` cannot run on Apple Silicon on this line: `frontend-maven-plugin` pins `nodeVersion v14.16.0`, and Node publishes no arm64 macOS build before 16, so the plugin fails with a 404 before reaching compilation. Worth knowing for whoever builds 8.7.1 on an M-series machine.


src/main/java/org/jahia/modules/sitesettings/users/UsersFlowHandler.java
```diff
@@ -216,6 +257,11 @@ public boolean removeUser(final String userKey, final MessageContext context) th
             @Override
             public Boolean doInJCR(JCRSessionWrapper session) throws RepositoryException {
                 JCRUserNode jahiaUser = userManagerService.lookupUserByPath(userKey);
+                if (jahiaUser == null || !isInAdministeredScope(jahiaUser.getPath())) {
+                    context.addMessage(new MessageBuilder().error().code(
+                            "siteSettings.user.remove.unsuccessful").arg(userKey).build());
```
baptistegrimaud: Out of curiosity — why is [#239](https://github.com/Jahia/siteSettings/pull/239) deliberately left out of this backport? I couldn't find the reasoning, and it happens to matter for this exact line.

On `main` this argument is `StringEscapeUtils.escapeXml(userKey)`, and [#239](https://github.com/Jahia/siteSettings/pull/239)'s own description names it directly — *"the user key reported when the path does not resolve"*. Without it, `8_7_x` reflects the raw request value:

- `removeUser`'s view-state is `model="userProperties"` and its `confirm` transition has no `bind="false"` and no `<binder>`, while `UserProperties.setUserKey(String)` is public — so `userKey` is request-bindable, not just whatever `populateUser` stored.
- Any value that isn't a real user path makes `lookupUserByPath` return `null`, which lands on exactly this branch with the supplied string in hand.
- `siteSettings.user.remove.unsuccessful=Couldn't remove user {0}.` interpolates it, and both views that can render the message print it raw — `removeUser.jsp:42` and `search.jsp:177` are bare `${message.text}`, no `<c:out>`, and JSP EL doesn't escape.

Pre-backport `removeUser` only ever passed `displayName` here, so the raw-value path arrives with this change rather than being inherited from the line.

**So either #239 comes along too, or — if it stays out of scope for a reason — just this one argument follows `main`.** That keeps the line byte-identical to `main`, which is the property the rest of the PR is careful to preserve, and needs `import org.apache.commons.lang.StringEscapeUtils;` back for its single use. Worth noting #239 explains why the encoder belongs on the argument rather than on `${message.text}`: several messages in this bundle intentionally carry markup.

**Source:** `main` escapes the same argument — [`UsersFlowHandler.java:265` @ `51744e0`](https://github.com/Jahia/siteSettings/blob/51744e0a47eaab043dc242c082ffdb41260554d3/src/main/java/org/jahia/modules/sitesettings/users/UsersFlowHandler.java#L265), introduced by [#239](https://github.com/Jahia/siteSettings/pull/239). Raw sinks on this branch: `src/main/resources/jnt_siteSettingsManageUsers/html/siteSettingsManageUsers.flow/removeUser.jsp:42` and `.../search.jsp:177`. Binding: `.../siteSettingsManageUsers.flow/flow.xml:56–62`. Bundle: `src/main/resources/resources/JahiaSiteSettings_en.properties:95`.

```suggestion
                    context.addMessage(new MessageBuilder().error().code(
                            "siteSettings.user.remove.unsuccessful").arg(StringEscapeUtils.escapeXml(userKey)).build());
```

.chachalog/B9.md
```diff
@@ -0,0 +1,5 @@
+---
+siteSettings: patch
+---
+
+Site-scoped administrators only manage principals within their site
```
baptistegrimaud: These three fragments have nothing to consume them on this line: `8_7_x` has no `.chachalog/` directory, none of the `chachalog-*.yml` workflows `main` carries, and `git grep -i chachalog` on the branch returns no hit at all. [#244](https://github.com/Jahia/siteSettings/pull/244), the precedent backport here, shipped none for the same reason.

Not harmful to keep them — they're inert either way. Just cleaner and more consistent with the line to leave them out, since chachalog isn't configured on this branch.

Worth being aware of the consequence though: nothing generates an 8.7.1 entry for this, and the behaviour boundary is the kind of thing an upgrader wants told — a deployment holding either screen in some other container goes from working to listing nothing and applying nothing. Might be worth a line wherever this line's release notes get written.

**Source:** `git ls-tree -r --name-only origin/8_7_x | grep -i chachalog` → empty; `main` has `.chachalog/` plus `.github/workflows/chachalog-comment-pr.yml` and `chachalog-prepare-changelog.yml`, neither present on `8_7_x`. Precedent: [#244](https://github.com/Jahia/siteSettings/pull/244) touched only the two JSPs.

src/main/java/org/jahia/modules/sitesettings/groups/ManageGroupsFlowHandler.java
```diff
@@ -55,11 +55,76 @@ public class ManageGroupsFlowHandler implements Serializable {
...
+
+    /**
+     * A target group or member principal is only in scope for the realm this management screen is bound
+     * to. When bound to a specific site (siteKey set from a jnt:virtualsite realm) the target must live
+     * under that site's principal tree (/sites/&lt;siteKey&gt;/...); in the server-wide realm (siteKey
+     * null) the target must live in the server-global store (/users/ or /groups/) and never inside a
+     * site's tree. Mirrors the store layout used by Jahia{User,Group}ManagerService. With no realm
+     * resolved at all nothing is in scope.
+     */
+    private boolean isInAdministeredScope(String principalPath) {
```
baptistegrimaud, line 87: I missed this during the review on `main`, but it looks like the two screens aren't symmetric on the read side — and since this backport carries the same code, it's worth writing down here.

`UsersFlowHandler` closes both halves: the write guards *and* the read paths (`search()` and `getProvidersList()` return empty when the realm is unresolved). `ManageGroupsFlowHandler` only got the write half — `search()` still calls `getGroupSearchResult(null, siteKey, …)` with `siteKey == null`, and the flow's `on-render` also calls `getProviders(false, false)`. So in an unresolved realm the writes correctly decline, but the screen still enumerates the **server-global** groups and providers.

Nothing to change in the backport — I diffed it and this is byte-identical to `main`, so the cherry-picks didn't introduce it. Two things follow though:

- The changelog fragment's "these screens list no principal" holds for Manage Users and is overstated for Manage Groups.
- Whoever re-tests this should know the group read side was never in scope, so an unresolved-realm listing showing global groups isn't a regression from this PR.

I'd rather follow up on `main` than widen this one — unless the read side was deliberately considered acceptable and I'm re-litigating a settled call?

**Source:** Branch `ManageGroupsFlowHandler.java:536–541` (`search()`) is byte-identical to `main`'s `:538–543`; the flow calls it plus `getProviders(false, false)` at `src/main/resources/jnt_siteSettingsManageGroups/html/siteSettingsManageGroups.flow/flow.xml:18–19`. Contrast `UsersFlowHandler.java:287` and `:431`, which do gate on `realmResolved`.

src/main/java/org/jahia/modules/sitesettings/users/UsersFlowHandler.java
eallais: Right on both counts, and it does arrive with the backport: pre-backport `removeUser` only ever passed `displayName`, so the `jahiaUser == null` branch is what puts the request value in the message. Applied your suggestion in [`6826bb1`](https://github.com/Jahia/siteSettings/commit/6826bb1), with the `StringEscapeUtils` import back — the line is byte-identical to `main` again.

On #239: no deliberate call against it. It simply isn't in what #248 asked for, and there is no backport request for it on this line. Its other escapes cover values `8_7_x` already reported raw before this backport — `displayName` in both `removeUser` messages — so they belong to a backport of #239 rather than riding along here. Happy to open one if you want it on 8.7.1.

Verification is `javac` over the module sources against the resolved dependency classpath, the same tier as the rest of this PR: `mvn` can't run on this line on Apple Silicon and there is no Cypress project here to exercise the message.

.chachalog/B9.md
eallais: Agreed — dropped in [`a39f9e1`](https://github.com/Jahia/siteSettings/commit/a39f9e1). Nothing on the line consumes them, and it matches what #244 carries.

Your consequence is the part worth keeping, so the behaviour boundary now lives in the PR description under a **Release notes** heading, as the line whoever writes the 8.7.1 notes can take verbatim. I narrowed its wording versus the fragment on `main`, for the reason in the other thread: only Manage Users lists no principal.

src/main/java/org/jahia/modules/sitesettings/groups/ManageGroupsFlowHandler.java
eallais: Not a settled call — you are not re-litigating anything. The read gating went into `UsersFlowHandler` and was never mirrored onto `ManageGroupsFlowHandler`; I missed it too.

Confirmed on both branches: `ManageGroupsFlowHandler.search()` calls `getGroupSearchResult(null, siteKey, …)` and the flow's `on-render` calls `getProviders(false, false)`, neither behind `realmResolved`, while `UsersFlowHandler:287` and `:431` are. So in an unresolved realm the group screen declines every write and still enumerates the server-global groups and providers.

Following your read: not widening this backport, follow-up on `main`. I have not filed it yet — I will, unless you would rather own it since you found it. Two things go in it: the gate on both read paths, and the fragment wording on `main`, which carries the same overstatement.

On this PR the wording point is moot — the fragments are gone (other thread) — and the release-notes line I put in the description is deliberately narrowed to "Manage Users additionally lists no principal". Whoever re-tests this can take an unresolved-realm listing of global groups as out of scope rather than as a regression.

baptistegrimaud: Took it rather than leaving it with you: #259 on `main`, and #260 for the maintenance line — stacked on this PR so it doesn't wait on `main`. It targets `audit/fix-settings-screen-realm-8_7_x` and GitHub retargets it to `8_7_x` once this merges. Both draft.

Both carry the same six guards. The cherry-pick onto this line conflicted twice, javadoc whitespace only, no code hunk. No `.chachalog/` fragment on #260, same reason as `a39f9e1`.

On your Apple-Silicon note — `mvn compiler:compile` invokes the compiler goal on its own, so the lifecycle never reaches `frontend-maven-plugin` and its `nodeVersion v14.16.0` 404. Enough to check Java on this line without a workaround; that's what I verified #260 with.

And the second item on your follow-up list dissolves rather than needing an edit: once #259 lands, `.chachalog/awPlC7Km.md`'s "these screens list no principal and apply no change" is simply true for both screens.

src/main/java/org/jahia/modules/sitesettings/users/UsersFlowHandler.java
eallais: Fixed in `6826bb16`: `arg(userKey)` → `arg(StringEscapeUtils.escapeXml(userKey))`, with `org.apache.commons.lang.StringEscapeUtils` imported back for that single use. The line is now byte-identical to `main`'s `UsersFlowHandler.java:265`.

On why #239 stays out: this backport is scoped to what #248 asked for, and #239's other `escapeXml` calls cover values `8_7_x` already reported raw before this change — they belong to a backport of #239, not to this one. Your argument is exactly the case that does not hold for: the raw-value path on this branch arrives *with* the scope commits, so that one argument follows `main`.

Verification: `Build Module`, `Sonar Analysis` and `Static Analysis` are green at head `a39f9e14`. No runtime check — `8_7_x` carries no `tests/` tree, so nothing on this line exercises the message.

.chachalog/B9.md
eallais: Fixed in `a39f9e14`: the three fragments (`B9.md`, `awPlC7Km.md`, `fW.md`) are dropped — nothing on `8_7_x` consumes them, same as #244.

On the consequence you flagged: since nothing generates an 8.7.1 entry for this, the release-note line is written into this PR's description under **Release notes**, deliberately narrower than the fragment's wording on `main` — Manage Users lists no principal, while Manage Groups' read side was never gated (see the other thread).

src/main/java/org/jahia/modules/sitesettings/groups/ManageGroupsFlowHandler.java
eallais: Confirmed, and you are not re-litigating a settled call — the read side was never gated, on this line or on `main`.

Tracked as a follow-up rather than widened into this PR: #259 gates `ManageGroupsFlowHandler`'s listing and lookup paths on the resolved realm on `main`, and #260 carries the same change for the 8.7.1 milestone — stacked on this branch, so it retargets to `8_7_x` once this one merges.

Your two consequences are recorded: this PR's description states the narrower boundary (Manage Users lists no principal; Manage Groups' read side out of scope), and it says explicitly that an unresolved-realm group listing is pre-existing behavior, not a regression of this backport.

eallais:
@baptistegrimaud all three review threads are answered and resolved:

- **`userKey` reported raw** — fixed in `6826bb16`: the argument now follows `main` (`escapeXml`, import restored). #239 as a whole stays out of scope, per the PR description.
- **`.chachalog/` fragments** — dropped in `a39f9e14`; the 8.7.1 release-note line lives in the PR description instead, worded for what actually holds on this line.
- **Manage Groups read side** — not widened into this backport; tracked by #259 on `main` and #260 for the 8.7.1 milestone.

CI is green at head `a39f9e14` (Build Module, Sonar Analysis, Static Analysis, module signature). Verification is build-level only: `8_7_x` carries no `tests/` tree, so nothing on this line exercises the change at runtime.

Ready for your review — #260 is stacked on this branch and retargets to `8_7_x` once this merges.

eallais:
@baptistegrimaud gentle ping on this one — it has been waiting since 04/08 and nothing blocks it any more: the five checks are green on `a39f9e14` (Build Module, Sonar Analysis, Static Analysis, signature, CLA) and all three review threads are answered and resolved. Could you take a look when you get a moment?

baptistegrimaud (approved):

PR #259, opened 2026-08-06
fix: list groups, providers and member candidates only for a realm the container resolves

baptistegrimaud:
## Summary

The Manage Groups administration screen manages the principals of one realm, the realm of the
container it is reached through: the global settings node carries the server-wide realm and its
server-global store, a site node carries that site's realm and store, and a container of any other
type carries no realm.

`ManageGroupsFlowHandler` records that resolution in `realmResolved`, and its write paths already
answer for the realm they are bound to. Its listing and lookup paths take the same rule here, so the
whole handler answers from one notion of what realm it manages — matching what `UsersFlowHandler`
already does for the Manage Users screen.

## Change

With no realm resolved, there is no store to read, so each read answers empty rather than the
server-global default a null `siteKey` selects:

- `search()` and `getProviders(…)` — the group list and the provider list of the screen's own view.
- `searchNewGroupMembers(…)` and `searchNewUserMembers(…)` — the candidate-member selectors.

Alongside that, three methods answer plainly for an input that resolves to no group:
`lookupGroupMembersCount(…)` returns `0`, and the two candidate-member methods return empty.

**Key resolution stays outside the realm bound**, deliberately. `lookupGroup(…)` resolves whatever key
it is given; the callers that act on the result carry the scope check, and the flow's own views expect
a resolved group. That is the same division of labour `UsersFlowHandler` uses — its `lookupUserByPath`
calls are likewise unbound, with `isInAdministeredScope` checking after. So what this PR bounds is
enumeration: what the screen lists without being told a key.

Every guard is on the no-realm state alone, so both containers the screen ships in — the site
settings page and the server settings page — behave exactly as they do today.

Both view skins (`siteSettingsManageGroups.flow` and
`siteSettingsManageGroups.settingsBootstrap3GoogleMaterialStyle.flow`) route through these same
handler methods, so one change covers both.

## Verification

`mvn compile` passes, and the full Cypress suite runs green in CI — including
`tests/cypress/e2e/Groups/manageGroupsRealm.test.cy.ts`, which drives all three containers through the
screen's own web flow.

`manageGroupsRealm`'s third case now also asserts what the no-realm container lists, read off the
response it already holds for its flow action.

That spec is what set the scope line above. An earlier revision of this branch bound key resolution
too; `manageGroupsRealm`'s third case went red, because `common/editMembersHead.jspf` renders
`user:displayName(group)` and the flow's views are written for a resolved group — six such expressions
across the two skins. Bounding enumeration and leaving resolution to the scope checks keeps the views
working and needs no change to any JSP.

Read surface enumerated from both `flow.xml` variants rather than from the handler alone: every
`flowHandler.*` expression either appears above, is bounded by a value the changes above bound
(`getSystemGroups` derives from `search()`), or already carries the realm (`initCriteria`,
`initGroup`).

`tests/cypress/e2e/Groups/manageGroupsRealm.test.cy.ts` covers the membership-write side of this
same realm binding, through the screen's own web flow, for all three containers. A listing-side case
is not part of this PR.

### DIFF
--- a/src/main/java/org/jahia/modules/sitesettings/groups/ManageGroupsFlowHandler.java
+++ b/src/main/java/org/jahia/modules/sitesettings/groups/ManageGroupsFlowHandler.java
@@ -598,6 +598,11 @@ public class ManageGroupsFlowHandler implements Serializable {
      * membership. Bounded by the realm the same way {@link #search()} is: with no realm resolved there
      * is no store to offer candidates from, so the result is empty.
      *
+     * The store searched is the one this screen administers — {@link #siteKey} — while the criteria
+     * supply the filters applied within it (searchIn, searchString, properties, storedOn, providers).
+     * Same split as {@code UsersFlowHandler.search(SearchCriteria)}, and the same one the membership
+     * flag below already uses.
+     *
      * @return the list of users, matching the specified search criteria
      */
     public Map<JahiaUser, Boolean> searchNewUserMembers(String groupKey, SearchCriteria searchCriteria) {
@@ -621,7 +626,7 @@ public class ManageGroupsFlowHandler implements Serializable {
 
         // Load all users, they will be displayed in the user selector for the group.
         Set<JCRUserNode> users = PrincipalViewHelper.getSearchResult(searchCriteria.getSearchIn(),
-                searchCriteria.getSiteKey(), searchCriteria.getSearchString(), searchCriteria.getProperties(), searchCriteria.getStoredOn(),
+                siteKey, searchCriteria.getSearchString(), searchCriteria.getProperties(), searchCriteria.getStoredOn(),
                 searchCriteria.getProviders());
 
         // Flag current group users
--- a/src/main/java/org/jahia/modules/sitesettings/groups/ManageGroupsFlowHandler.java
+++ b/src/main/java/org/jahia/modules/sitesettings/groups/ManageGroupsFlowHandler.java
@@ -292,11 +292,16 @@ public class ManageGroupsFlowHandler implements Serializable {
     }
 
     /**
-     * Returns a map of all group providers currently registered.
+     * Returns the principal providers mounted for the administered realm. Bounded the same way the
+     * listing is: with no realm resolved there is no realm whose providers to enumerate, so the result
+     * is empty rather than the server-global set a null {@link #siteKey} would otherwise select.
      *
-     * @return a map of all group providers currently registered
+     * @return the providers of the administered realm
      */
     public Set<String> getProviders(final boolean isUsers, final boolean includeGlobals) throws RepositoryException {
+        if (!realmResolved) {
+            return Collections.emptySet();
+        }
         return JCRTemplate.getInstance().doExecuteWithSystemSession(new JCRCallback<Set<String>>() {
             @Override
             public Set<String> doInJCR(JCRSessionWrapper session) throws RepositoryException {
@@ -375,24 +380,28 @@ public class ManageGroupsFlowHandler implements Serializable {
     }
 
     /**
-     * Looks up the specified group by key.
+     * Looks up the specified group by key, within the administered realm. With no realm resolved there is
+     * no store to look the key up in, so the answer is none.
      *
      * @param selectedGroup
      *            the group key
      * @return up the specified group by key
      */
     public JCRGroupNode lookupGroup(String selectedGroup) {
-        return selectedGroup != null ? groupManagerService.lookupGroupByPath(selectedGroup) : null;
+        return realmResolved && selectedGroup != null ? groupManagerService.lookupGroupByPath(selectedGroup) : null;
     }
 
     /**
      * Return the count of members for a group, only working on jcr groups
      * if an external group is used it will return 0
      * @param group the group to get the members from
-     * @return the count of members for a group, 0 if group is external
+     * @return the count of members for a group, 0 if group is external or none was resolved
      * @throws RepositoryException
      */
     public long lookupGroupMembersCount(JCRGroupNode group) throws RepositoryException {
+        if (group == null) {
+            return 0;
+        }
         if (!group.getProperty("j:external").getBoolean()) {
             QueryResult result = group.getSession().getWorkspace().getQueryManager()
                     .createQuery("select * from [jnt:member] as member where " +
@@ -531,11 +540,17 @@ public class ManageGroupsFlowHandler implements Serializable {
     }
 
     /**
-     * Performs the group search.
+     * Lists the groups of the administered realm. The realm is what bounds the listing: a site realm
+     * lists that site's store, the server-wide realm the server-global one. With no realm resolved there
+     * is no store to list, so the result is empty rather than the server-global default a null
+     * {@link #siteKey} would otherwise select.
      *
      * @return the list of groups, matching the specified search criteria
      */
     public Set<JCRGroupNode> search() {
+        if (!realmResolved) {
+            return Collections.emptySet();
+        }
         long timer = System.currentTimeMillis();
         Set<JCRGroupNode> searchResult = PrincipalViewHelper.getGroupSearchResult(null, siteKey, null, null, null, null, false);
         logger.info("Found {} groups in {} ms", searchResult.size(), System.currentTimeMillis() - timer);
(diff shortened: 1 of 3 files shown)


src/main/java/org/jahia/modules/sitesettings/groups/ManageGroupsFlowHandler.java
```diff
@@ -531,22 +541,34 @@ public void removeMembers(String groupKey, String[] members, MessageContext cont
...
-     * Performs the group search.
+     * Lists the groups of the administered realm. The realm is what bounds the listing: a site realm
+     * lists that site's store, the server-wide realm the server-global one. With no realm resolved there
+     * is no store to list, so the result is empty rather than the server-global default a null
+     * {@link #siteKey} would otherwise select.
      *
      * @return the list of groups, matching the specified search criteria
      */
     public Set<JCRGroupNode> search() {
+        if (!realmResolved) {
```
eallais, line 552: Nothing in the suite exercises any of the four guards, so reverting one of them ships green. The Users side landed its listing cases together with the guard (`manageUsersRealm`), and here `manageGroupsRealm`'s third case is already holding the response that would assert it.

- That case does `cy.request({method: 'GET', url: render})` on the no-realm container and keeps the body to dig the flow action out of it. The listing is in that same body, so `expect(res.body).not.to.contain(noRealmGroup)` is the whole assertion — and it is non-vacuous: `noRealmGroup` is created in the server-global store, which is exactly what a null `siteKey` used to select.
- The two realm cases in that spec stay the positive controls, unchanged.

**Worth adding before this goes ready — it's the only thing keeping the four guards from quietly rotting.**

<sub>Anchor for the assertion: `tests/cypress/e2e/Groups/manageGroupsRealm.test.cy.ts:151-160`, the `cy.request` GET whose body already reaches that point.</sub>

src/main/java/org/jahia/modules/sitesettings/groups/ManageGroupsFlowHandler.java
```diff
@@ -389,10 +396,13 @@ public JCRGroupNode lookupGroup(String selectedGroup) {
      * Return the count of members for a group, only working on jcr groups
      * if an external group is used it will return 0
      * @param group the group to get the members from
-     * @return the count of members for a group, 0 if group is external
+     * @return the count of members for a group, 0 if group is external or none was resolved
      * @throws RepositoryException
      */
     public long lookupGroupMembersCount(JCRGroupNode group) throws RepositoryException {
+        if (group == null) {
```
eallais, line 403: These three null branches — here, and the `groupNode == null` ones in `searchNewGroupMembers` / `searchNewUserMembers` — don't rescue the request: the views dereference the same null a moment later.

- `editGroup.jsp:84` and `common/editMembersHead.jspf:14` both render `user:displayName(group)`, and `PrincipalViewHelper.getDisplayName(null)` falls through to its `else` branch, where building the `IllegalArgumentException` message calls `p.getClass()` — NPE.
- So for a key that resolves to nothing the response is a 500 either way: before, from `group.getProperty("j:external")`; after, one line later from the JSP.

Either guard the views too (`<c:if test="${not empty group}">`) or drop the branches — as they stand they read like robustness the request never gets.

<sub>`org/jahia/data/viewhelper/principal/PrincipalViewHelper.java:267-292` — `getDisplayName(Object, Locale)` has no null branch; the final `else` throws with `p.getClass().getName()` in the message.</sub>

eallais (changes requested):

src/main/java/org/jahia/modules/sitesettings/groups/ManageGroupsFlowHandler.java
```diff
@@ -571,11 +593,24 @@ public int compare(JCRNodeWrapper o1, JCRNodeWrapper o2) {
...
 
+    /**
+     * Lists the users of the administered realm as candidate members, flagged with their current
+     * membership. Bounded by the realm the same way {@link #search()} is: with no realm resolved there
+     * is no store to offer candidates from, so the result is empty.
+     *
+     * @return the list of users, matching the specified search criteria
+     */
     public Map<JahiaUser, Boolean> searchNewUserMembers(String groupKey, SearchCriteria searchCriteria) {
+        if (!realmResolved) {
```
eallais, line 604: `searchNewUserMembers` builds its store from `searchCriteria.getSiteKey()` (line 623), not from the handler's own `siteKey` — so the method does not answer for the realm this guard just checked, and the new docstring's "users of the administered realm" does not hold.

Same shape as `UsersFlowHandler.search(SearchCriteria)`: keep the criteria for `searchIn` / `searchString` / `properties` / `storedOn` / `providers`, and pass the handler's `siteKey`.

```java
Set<JCRUserNode> users = PrincipalViewHelper.getSearchResult(searchCriteria.getSearchIn(),
        siteKey, searchCriteria.getSearchString(), searchCriteria.getProperties(), searchCriteria.getStoredOn(),
        searchCriteria.getProviders());
```

<sub>Precedent: `src/main/java/org/jahia/modules/sitesettings/users/UsersFlowHandler.java:301`.</sub>

src/main/java/org/jahia/modules/sitesettings/groups/ManageGroupsFlowHandler.java
eallais: Correcting my own premise — it only holds for `lookupGroupMembersCount`. `flow.xml:61-62` evaluates `lookupGroup` then the count, and `editGroup.jsp` dereferences the same null right after, so there the branch changes nothing the request sees.

For `searchNewGroupMembers` / `searchNewUserMembers` it does change something: on `main` the search runs *before* the dereference (`PrincipalViewHelper.getSearchResult(…)`, then `groupNode.getMembers()`), so the new branches stop a full-store search from running for a key that resolves to nothing. Worth keeping as they are.

Downgrading this to a nit — no JSP change needed in this diff. The `user:displayName(null)` 500 (`PrincipalViewHelper.getDisplayName` has no null branch) deserves its own ordinary bug report instead.

src/main/java/org/jahia/modules/sitesettings/groups/ManageGroupsFlowHandler.java
baptistegrimaud: Agreed, and added in `255bb7d` — read off the same response, no extra request.

I checked the one thing that would have made it vacuous: truncation. `manageGroups.jsp:144` is a bare `<c:forEach items="${groups}">` with no `end=`, and line 149 renders the display name, so a pre-fix run would genuinely have contained that unique name. Your non-vacuity argument holds.

One narrowing on scope, so nobody reads more into it later: this asserts `search()`. `searchNewGroupMembers` / `searchNewUserMembers` sit behind `editUsersMembers` / `editGroupsMembers` transitions the GET never reaches, so those two can still rot silently — and `getProviders` only shows up indirectly. Happy to add cases for them if you think it's worth the fixtures; I left them out to keep this diff to the assertion you asked for.

src/main/java/org/jahia/modules/sitesettings/groups/ManageGroupsFlowHandler.java
baptistegrimaud: Your correction is the accurate version, and it matches what I read: `flow.xml:61-62` evaluates the count and then `editGroup.jsp:84` dereferences the same null regardless, so the `lookupGroupMembersCount` branch changes nothing the request sees — while in the two search methods `PrincipalViewHelper.getSearchResult(…)` runs before the dereference, so those branches do stop a full-store search.

Leaving all three as they are, and the docstring no longer implies the count branch rescues anything.

Agreed the `user:displayName(null)` 500 wants its own ordinary report rather than riding along here — it's a crash on a malformed key, it discloses nothing, and it predates this change. Say the word if you'd rather I open it than you.

src/main/java/org/jahia/modules/sitesettings/groups/ManageGroupsFlowHandler.java
baptistegrimaud: Good catch, and you're right that the store selection and the realm gate are two different things.

On reflection this doesn't belong in this PR: it isn't the no-realm state this branch is about, it needs its own reasoning about which inputs the criteria are entitled to choose, and folding it in here would have shipped it with no regression cover of its own. So I've backed it out — `8f763b5`, cherry-picked to #260 as `cf984d0` — and it's tracked separately on the private side, where the analysis can live in full.

What stays here is the docstring: it no longer claims this method answers for the administered realm, since the guard it carries only covers the no-realm state.

Happy to take the follow-up myself unless you'd rather own it.

PR #262, opened 2026-08-06
fix: render settings components only for callers administering the resource

AKarmanov:
## Summary

Ports the development line's condition to this maintenance line. The permission requirement belongs on the
**component**, not only on the settings template that normally hosts it (`j:requiredPermissionNames`): a
component's access rule should travel with the component and hold on every render path, regardless of where
it is placed.

## Change

A render filter (`SettingsComponentPermissionFilter`, priority 21.5) applied on the settings node types this
line actually ships, registered via OSGi Declarative Services. When the caller holds none of the accepted
permissions it returns an empty fragment — core's own semantics in `TemplatePermissionCheckFilter`. Because
`WebflowAction` re-enters the render chain for each webflow POST, every transition is covered too, not just
the initial GET.

**The check is evaluated on the render's MAIN RESOURCE, not on the component node.** This is the
counter-intuitive part and it is load-bearing: a legitimate screen's component node lives inside its module
(`/modules/...`), where a site-scoped administrator holds nothing — so the obvious component-node check
would refuse real administrators.

The accepted permissions are the coarse core ones (`root-permissions.xml`). The finer per-screen permissions
come from a module's own `permissions.xml` and resolve to `false` indistinguishably from a denial where they
are not registered, which would fail closed for administrators too; the finer requirement still applies on
the administration route via the template.

## Verification — what was and was not done

The mechanism is verified on the development line, where the identical filter was measured registering at
priority 21.5 and refusing a low-privilege and an anonymous caller across every gated node type, with read
controls confirming the negatives are permission decisions rather than unreadable pages, and administrators
unaffected (the no-over-blocking control).

**Not verified on this line** — not built or deployed here. The functional check is one render of a settings
screen as an administrator (must still work) and as an ordinary editor (must render nothing).

### DIFF
--- a/src/main/java/org/jahia/modules/sitesettings/render/SettingsComponentPermissionFilter.java
+++ b/src/main/java/org/jahia/modules/sitesettings/render/SettingsComponentPermissionFilter.java
@@ -1,0 +1,118 @@
+/*
+ * Copyright (C) 2002-2022 Jahia Solutions Group SA. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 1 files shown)


eallais:
Closing for the same reason as #263: `jahia-pack-private` main pins `siteSettings` to `8.13.0-SNAPSHOT` ([Jahia/jahia-pack-private#1059](https://github.com/Jahia/jahia-pack-private/pull/1059)), the next release is 8.13.0 from `main` (which carries #240), and no 8.12.4 release is planned from the 8.12 maintenance line. The branch stays in place, so this can be reopened if a consumer turns out to be pinned to an 8.12.x release.

PR #263, opened 2026-08-06
fix: scope the user and group management screens to the realm they administer

eallais:
Backport onto `8_12_y` — the live 8.12 line (`8.12.4-SNAPSHOT`; `8_12_x` stopped at `8.12.0`) — of the two commits from #248.

Both Cypress specs travel: unlike `8_7_x`, this line carries a `tests/` project and a `.chachalog/` directory, so the specs and the changelog fragment come along unchanged.

## What is in it

`8_12_y` already carries the earlier scope work — `isInAdministeredScope`, `isAllowedMember` and `isSiteKeyInScope` are present in both handlers, and the five commits from #226 / #227 are ancestors of this branch. So only #248's own two commits are needed:

| commit | what it does |
|---|---|
| `27dd2c8` | bind the user and group management screens to the realm of their container |
| `ead9725` | list principals only for a realm the container resolves |

The Java applied without conflict on both handlers. The only conflict was the Cypress layout: `main` reorganized `tests/cypress/e2e/` into per-screen subdirectories (`Groups/`, `Users/`), while this line keeps a flat tree. The two specs therefore land at `tests/cypress/e2e/manageGroupsRealm.test.cy.ts` and `tests/cypress/e2e/manageUsersRealm.test.cy.ts`, with their two relative imports adjusted from `../../` to `../` to match that depth. Their content is otherwise untouched.

Everything the specs rely on exists on this line: `@jahia/cypress` is `^3.4.0` on both sides, `cy.executeGroovy` is already used by three specs here, and `SiteSettingsGroups.visit` / `SiteSettingsUsers.visit` are present in the page objects.

## How close is this to `main`

`ManageGroupsFlowHandler` is now **byte-identical** to `main`.

`UsersFlowHandler` differs by 7 insertions / 11 deletions, and all of it is pre-existing divergence of the line, unrelated to this change: `main` passes message arguments through `StringEscapeUtils.escapeXml` in `bulkCreateUsers` and `removeUser`, which came with #239 and was never brought to this line. Nothing here introduces those call sites — the `siteSettings.user.remove.unsuccessful` message on an unresolved path already exists on `8_12_y` at its own line 238, guarded by the `isInAdministeredScope` check that arrived with #226. Bringing the encoders over belongs to a backport of #239, not to this one.

## Verification

`mvn test-compile` on this branch: **BUILD SUCCESS** (21 s), both handler classes produced. The two `non-varargs call of varargs method` warnings on `UsersFlowHandler` are pre-existing on the line — this change adds no `.arg(...)` / `.args(...)` call site, and the warnings only shift by the lines added above them.

The two specs have not yet been run against a deployed `8.12.4-SNAPSHOT` bundle; on `main` they cover exactly the three cases they assert here (both realms mutate and list, a container carrying no realm does neither), with the two realm cases acting as the positive controls for the third.

### DIFF
--- a/.chachalog/awPlC7Km.md
+++ b/.chachalog/awPlC7Km.md
@@ -2,4 +2,4 @@
 siteSettings: patch
 ---
 
-Aligned the Manage Users and Manage Groups administration screens with the principal realm of the container they are reached through: the server-wide realm on the global settings node, a single site's realm on a site node. Held by any other container, these screens list principals and apply no change.
+Aligned the Manage Users and Manage Groups administration screens with the principal realm of the container they are reached through: the server-wide realm on the global settings node, a single site's realm on a site node. Held by any other container, these screens list no principal and apply no change.
--- a/.chachalog/awPlC7Km.md
+++ b/.chachalog/awPlC7Km.md
@@ -1,0 +1,5 @@
+---
+siteSettings: patch
+---
+
+Aligned the Manage Users and Manage Groups administration screens with the principal realm of the container they are reached through: the server-wide realm on the global settings node, a single site's realm on a site node. Held by any other container, these screens list principals and apply no change.
--- a/src/main/java/org/jahia/modules/sitesettings/groups/ManageGroupsFlowHandler.java
+++ b/src/main/java/org/jahia/modules/sitesettings/groups/ManageGroupsFlowHandler.java
@@ -55,10 +55,24 @@ public class ManageGroupsFlowHandler implements Serializable {
 
     private String siteKey;
 
+    private boolean realmResolved;
+
+    /**
+     * Resolves the principal realm this screen manages from the container it is reached through: a site
+     * node scopes it to that site's principals, the global settings node to the server-global store
+     * (hence a null {@link #siteKey}). Any other container carries no realm, leaving
+     * {@link #realmResolved} false — the state in which every scope check below answers false.
+     */
     public void initRealm(RenderContext renderContext) throws RepositoryException {
         JCRNodeWrapper mainNode = renderContext.getMainResource().getNode();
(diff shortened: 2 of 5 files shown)


baptistegrimaud:
@eallais : my understanding is that the upcoming release of siteSettings will be 8.13.0 (there shouldn't be any 8.12.4 release from 8.12.x maintenance branch). 8.13.0 should include https://github.com/Jahia/siteSettings/pull/248.
So I guess this PR can be closed and not merged? But maybe I missed something?

eallais:
@baptistegrimaud You're right, nothing missed — closing it.

Confirmed on the delivery side: `jahia-pack-private` main has pinned `siteSettings` to `8.13.0-SNAPSHOT` since [Jahia/jahia-pack-private#1059](https://github.com/Jahia/jahia-pack-private/pull/1059) (2026-08-04), so the pack under test takes `main`, which carries #248 (merged as `2bd4aef`). And `8_12_y` cut `8.12.3` on 2026-08-05, one day before this PR existed — it could only ever have shipped in an `8.12.4` that isn't planned, and there is no milestone for it.

No exposure stays behind either: `8.13.0-SNAPSHOT` declares the same parent (`jahia-modules 8.2.2.0`) and the same `jahia-depends default` as `8_12_y`, so anyone on the 8.12 line can take 8.13.0 without a Jahia upgrade. The maintenance line that genuinely owed a backport was `8_7_x`, delivered by #249.

Same reasoning applies to the two other open 8.12 backports of this audit — #262 (`8_12_y`) and #261 (`8_12_x`, a line that stopped at `8.12.0`). I am closing those as well; their branches stay in place, so reopening is a one-click matter if you find a consumer still pinned to an 8.12.x release.

